PR #4, opened 2020-10-30
remove prefix from list of columns - from hubspot

fivetran-jamie:


macros/remove_prefix_from_columns.sql
```diff
@@ -0,0 +1,12 @@
+{% macro remove_prefix_from_columns(columns, prefix='property_', exclude=[]) %}
```
fivetran-kristin, line 1: For this general package, I would remove the default prefix parameter value. 
```suggestion
{% macro remove_prefix_from_columns(columns, prefix='', exclude=[]) %}
```

fivetran-kristin (approved):
Looks good, I just have one suggestion for your review

macros/remove_prefix_from_columns.sql
fivetran-jamie: sounds good -- i'll change it here and add `prefix='property_'` to any references to this macro (it's just used in the hubspot src package i believe)

fivetran-kristin: FYI this will have implications for the current draft of the hubspot PRs. When using this macro, you will need to explicitly define the prefix value.

fivetran-kristin: (whoops, did not see your previous comment!) You're on it 💪

PR #6, opened 2020-11-17
Quote columns in fill columns macro

DylanBaker:


DylanBaker:
@fivetran-kristin, I'm updating this macro so that we can better handle reserved words, i.e. 'from' in the Hubspot package.

macros/fill_staging_columns.sql
```diff
@@ -13,4 +13,16 @@
     {%- if not loop.last -%} , {% endif -%}
 {% endfor %}
 
-{% endmacro %}
\ No newline at end of file
+{% endmacro %}
+
+{% macro quote_column(column) %}
+    {% if 'quote' in column %}
```
fivetran-jamie, line 19: so `quote` would be an optional boolean argument in the `staging_columns` macro array? @DylanBaker

DylanBaker: Yeah, it would look something like this (from the proposed hubspot changes):
```
{% macro get_email_event_dropped_columns() %}

{% set columns = [
    {"name": "_fivetran_synced", "datatype": dbt_utils.type_timestamp()},
    {"name": "bcc", "datatype": dbt_utils.type_string()},
    {"name": "cc", "datatype": dbt_utils.type_string()},
    {"name": "drop_message", "datatype": dbt_utils.type_string()},
    {"name": "drop_reason", "datatype": dbt_utils.type_string()},
    {"name": "id", "datatype": dbt_utils.type_string()},
    {"name": "reply_to", "datatype": dbt_utils.type_string()},
    {"name": "subject", "datatype": dbt_utils.type_string()}
] %}

{% if target.type == 'snowflake' %}
 {{ columns.append({"name": "FROM", "datatype": dbt_utils.type_string(), "quote": True, "alias": "from_email"}) }}
{% else %}
 {{ columns.append({"name": "from", "datatype": dbt_utils.type_string(), "quote": True, "alias": "from_email"}) }}
{% endif %}

{{ return(columns) }}

{% endmacro %}

```

fivetran-jamie (approved):
Looks good to me!

PR #7, opened 2020-11-23
remove /n from the tmp models

fivetran-jamie:
little fix -- currently the tmp models are initialized as:
`select *\nfrom {{ var('table_name') }}`

which requires you to go in and delete the problematic \n

### DIFF
--- a/columns_setup.sh
+++ b/columns_setup.sh
@@ -2,7 +2,7 @@
 mkdir -p $1/macros
 mkdir -p $1/models/tmp
 dbt run-operation fivetran_utils.generate_columns_macro --args '{"table_name": "'$5'", "schema_name": "'$4'", "database_name":"'$3'"}' | tail -n +2 > $1/macros/get_$5_columns.sql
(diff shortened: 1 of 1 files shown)


columns_setup.sh
```diff
@@ -2,7 +2,7 @@
 mkdir -p $1/macros
 mkdir -p $1/models/tmp
 dbt run-operation fivetran_utils.generate_columns_macro --args '{"table_name": "'$5'", "schema_name": "'$4'", "database_name":"'$3'"}' | tail -n +2 > $1/macros/get_$5_columns.sql
-echo "select *\nfrom {{ var('$5') }}" > $1/models/tmp/$2__$5_tmp.sql
```
DylanBaker, line 5: @fivetran-jamie So, in theory, that should be adding a new line, not be printed out explicitly as it is here. It does so for me, but I'm wondering if there's a difference between how different command lines work. 

Which command line are you using?

All that being said, I'm totally happy for this change. It's really just an aesthetics thing by having it on a new line.

DylanBaker (approved):

PR #10, opened 2020-12-09
Median and first_value macros

fivetran-joemarkiewicz:
- Added a macro to calculate the median of a field with cross db functionality.
- Added a first_value macro to determine the first (or last) value with cross db functionality.
- Updated ReadMe with documentation on above two macros.


macros/first_value.sql
```diff
@@ -0,0 +1,19 @@
+{% macro first_value(first_value_field, partition_field, order_by_field, order="asc") -%}
+
+{{ adapter.dispatch('first_value', packages = fivetran_utils._get_utils_namespaces()) (first_value_field, partition_field, order_by_field, order="asc") }}
+
+{%- endmacro %}
+
+--Default first_value calculation
+{% macro default__first_value(first_value_field, partition_field, order_by_field, order="asc")  %}
+
+    first_value( {{ first_value_field }} ) over (partition by {{ partition_field }} order by {{ order_by_field }} {{ order }} )
```
fivetran-jamie, line 10: @fivetran-joemarkiewicz i think `first_value` by default includes nulls -- perhaps we should include that as an argument to pass this macro?

macros/first_value.sql
```diff
@@ -0,0 +1,19 @@
...
+
+{% endmacro %}
+
+--first_value calculation specific to Redshift
+{% macro redshift__first_value(first_value_field, partition_field, order_by_field, order="asc") %}
+
+    first_value( {{ first_value_field }} ) over (partition by {{ partition_field }} order by {{ order_by_field }} {{ order }} , {{ partition_field }} rows unbounded preceding )
```
fivetran-jamie, line 17: is ordering by the partition_field necessary in redshift?

macros/first_value.sql
```diff
@@ -0,0 +1,19 @@
+{% macro first_value(first_value_field, partition_field, order_by_field, order="asc") -%}
+
+{{ adapter.dispatch('first_value', packages = fivetran_utils._get_utils_namespaces()) (first_value_field, partition_field, order_by_field, order="asc") }}
```
fivetran-jamie, line 3: should `order="asc"` be included in the dispatch call?

README.md
```diff
@@ -44,6 +55,17 @@ Usage:
...
+This macro is used to return the median value of a field with cross db functionality. The default median calulcation within the macro is the `median` function. The BigQuery median calculation within the macro is the `percentile_cont` function.
+
+**Usage:**
+```sql
+{{ fivetran_utils.median(median_field='time_to_close', partition_field='id', percent=0.5) }}
+```
+**Args:**
+* `median_field` (required): Name of the field of which you are determining the median value.
+* `partition_field` (required): Name of the field you want to partition by to determine the median value.
+* `percent` (default = optional, bigquery = required): The percent necessary for `percentile_cont` to determine the median value. By defualt the value is `0.5`.
```
fivetran-jamie, line 68: is there a case in which someone would use a different percent to calculate the median? if not, might be simpler to remove this argument and just always use 0.5 🤷

fivetran-jamie (commented):
looks mostly good but left a couple of comments!

macros/first_value.sql
fivetran-joemarkiewicz: It does include null values by default. I included the argument to the macros as we do want to ignore nulls for this output. Thanks!

macros/first_value.sql
fivetran-joemarkiewicz: Strangely yes. When I remove the partition by statement I receive the first value for for the first conversation (in Intercom terms) instead of the first value for each individual conversation. I agree it does not seem necessary to include, but does not provide reliable results when removed.

macros/first_value.sql
fivetran-joemarkiewicz: Nope, removed!

README.md
fivetran-joemarkiewicz: Completely changed this macro to instead be a percentile macro which we can use for median if we want.

fivetran-jamie (approved):
LGTM!

PR #13, opened 2021-02-09
add enabled_vars_one_true

fivetran-joemarkiewicz:
This branch includes the following updates:
- Addition of the macro `enabled_vars_one_true` and accompanying documentation.

The `enabled_vars_one_true` macro is used in the [quickbooks sales_union intermediate model](https://github.com/fivetran/dbt_quickbooks/blob/sales_receipt_var/models/intermediate/int_quickbooks__sales_union.sql). The purpose of this macro is to run the model if one of the variables passed = `true`. If they all are false then the macro returns false and the model does not run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,18 @@ This macro creates a dummy coalesce value based on the data type of the field. S
 * `column` (required): Field you are applying the dummy coalesce.
 
 ----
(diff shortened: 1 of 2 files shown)


fivetran-kristin (approved):
LGTM -- agreed that the name isn't great, but I can't think of anything better so this will do!

PR #15, opened 2021-02-25
more quoting safeguards

fivetran-jamie:


fivetran-jamie:
this is working for greenhouse but if someone could just approve as a sanity check 😅

fivetran-joemarkiewicz (approved):
Just tested it on Zendesk and Github packages and the updates worked! LGTM

PR #17, opened 2021-03-10
Max bool macro

fivetran-joemarkiewicz:
This pull request includes the addition of the max_bool macro. This macro allows the user to return the max boolean of a field for Snowflake, Redshift, BigQuery, and Postgres.


fivetran-jamie (approved):
LGTM

PR #19, opened 2021-03-22
Staging helper macro

fivetran-joemarkiewicz:
This PR includes the following changes:

- Addition of the `staging_models_automation` macro which allows for multiple bash commands to be generated within the CLI to create staging models/macros. This CLI output can then be copy/pasted to generate all necessary staging models/macros.
- Inclusion of accompanying `staging_models_automation` documentation.
- Addition of the `PULL_REQUEST_TEMPLATE.md` to be used when merged with master for the dbt_fivetran_utils PR template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ This macro is intended to be used as a `run-operation` when generating Fivetran 
 
 **Usage:**
 ```bash
-dbt run-operation staging_models_automation --args '{package: asana, target_schema: asana_target, target_database: database-target-name, tables: ["user","tag"]}'
+dbt run-operation staging_models_automation --args '{package: asana, source_schema: asana_source, source_database: database-source-name, tables: ["user","tag"]}'
 ```
 
 **CLI Output:**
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -228,6 +228,27 @@ It simply chooses which version of the data to seed (the Snowflake copy should c
 **Args:**
 * `seed_name` (required): Name of the seed that has separate snowflake seed data.
 
+----
+### staging_models_automation ([source](macros/staging_models_automation.sql))
+This macro is intended to be used as a `run-operation` when generating Fivetran dbt source package staging models/macros. This macro will receive user input to create all necessary ([bash commands](columns_setup.sh)) appended with `&&` so they may all be ran at once. The output of this macro within the CLI will then be copied and pasted as a command to generate the staging models/macros.
+
+**Usage:**
+```bash
+dbt run-operation staging_models_automation --args '{package: asana, target_schema: asana_target, target_database: database-target-name, tables: ["user","tag"]}'
```
fivetran-jamie, line 237: got a little confused by the use of `target_*`. in this case, the target_schema is the source schema we're reading from right? and not the dbt target.schema we're writing these models to? if so, maybe we'd wanna use a different naming convention

fivetran-jamie: or we could add some more detailed description to the args section below

fivetran-joemarkiewicz: Nope you are completely right! I meant to use source instead of target for the schema and database, thanks I will update it to reflect source!

fivetran-joemarkiewicz: Updated the macro and readme to reflect source opposed to target.

fivetran-jamie (approved):

PR #20, opened 2021-03-24
Dynamic seed macro

fivetran-joemarkiewicz:
While testing the Zendesk package for postgres compatibility I found a similar issue with seed data where a new data file was required. In order to accomodate for this I created a replica of the `snowflake_seed_data` macro with the addition of the `warehouse` argument which allows the user to specify the warehouse we need a new seed file for.

This was tested within the Zendesk integration_tests folder and worked for both postgres and snowflake seed data files. My thought is to keep 'snowflake_seed_data' for now since a lot of packages utilize the macro, but use the `seed_data_helper` macro for future packages and updates.


README.md
```diff
@@ -228,6 +228,21 @@ It simply chooses which version of the data to seed (the Snowflake copy should c
...
+----
+### seed_data_helper ([source](macros/seed_data_helper.sql))
+This macro is intended to be used when a source table column is a reserved keyword in a warehouse, and Circle CI is throwing a fit.
+It simply chooses which version of the data to seed.
+
+***Usage:**
+```yml
+    # in integration_tests/dbt_project.yml
+    vars:
+        table_name: "{{ fivetran_utils.seed_data_helper(seed_name='user_data', warehouse='postgres') }}"
```
fivetran-jamie, line 240: would we be able to use this if there was a reserved keyword in both snowflake and postgres?

fivetran-jamie: not sure if having the warehouse as an argument works here

fivetran-joemarkiewicz: This would not work if a reserved word was in both snowflake and postgres. That is a good point and something this macro should be able to handle. 

I have updated the macro to allow the `warehouses` argument to be a list and you may pass in both snowflake and postgres which the macro can then capture the snowflake and postgres specific seed data if needed!

fivetran-jamie (approved):
looks good to me!

macros/seed_data_helper.sql
```diff
@@ -0,0 +1,13 @@
+{% macro seed_data_helper(seed_name, warehouses) %}
+
+{% if target.type in warehouses %}
+    {% for w in warehouses %}
+        {% if target.type == w %}
+            {{ return(ref(seed_name ~ "_" ~ w ~ "")) }}
```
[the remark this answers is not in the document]
fivetran-joemarkiewicz: I came across a weird *bug* where I would see the macro fail without it? I am not entirely sure if it is needed, but felt it was safer to include in case the *bug* appeared again.

PR #21, opened 2021-03-25
update namespace var to fivetran_utils_dispatch_lis

fivetran-kristin:
This change is necessary for spark utils compatibility as outlined in [this doc.](https://docs.google.com/document/d/1TpCxX8XW8W4Au-dsvsZgsffTuTvyEnjMMe3WHW9Pdo0/edit#)


fivetran-joemarkiewicz (approved):
The update looks good! 

I am curious what this change does for Spark compatibility? Also, is it possible for this update to effect our other packages? I found within our [dbt_zendesk](https://github.com/fivetran/dbt_zendesk/blob/master/dbt_project.yml#L54) package we declare `dbt_utils_dispatch_list`. Would we want to update this to instead be `fivetran_utils_dispatch_list`?

PR #22, opened 2021-03-30
add pass through macro

fivetran-joemarkiewicz:

The addition of the `add_pass_through_columns` macro. This macro allows a user to declare a pass through variable in the following format:
```
dbt_project.yml

...

vars:

  deal_pass_through_columns:
  
        - name: "property_sales_development_representative"
          alias: "sales_development_representative_owner_id"
  
        - name: "property_energy_consultant"
          alias: "energy_consultant_owner_id"

```

This macro was tested within the dbt_hubspot_source package and was originally created by @jamesrayoub within the respective [PR](https://github.com/fivetran/dbt_hubspot_source/pull/35)

This will only be used within the dbt_hubspot_source package. However, it can be applied to other packages that accept pass through fields as variables.

- [x] Yes


fivetran-joemarkiewicz:
An additional macro was added to allow for the code within the staging model to remain dry and dynamic. The `fill_pass_through_columns` macro was tested and documented.

macros/add_pass_through_columns.sql
```diff
@@ -0,0 +1,13 @@
+{% macro add_pass_through_columns(base_columns, pass_through_var) %}
+
+  {% if pass_through_var %}
+
+    {% for column in pass_through_var %}
+
+      {% do base_columns.append({ "name": column.name, "datatype": column.datatype, "alias": column.alias }) if column.alias else base_columns.append({ "name": column.name, "datatype": column.datatype}) %}
```
fivetran-jamie, line 7: are we removing the datatype attribute from this part? forgot where we left off

fivetran-joemarkiewicz: Yeah I think it makes the most sense to remove it as it is not needed and adds unnecessary complexity. Removed, thanks!

fivetran-jamie (approved):
looks good! i would just maybe add some more documentation for the macros (like in what types of files they're called in) but the code is all good

PR #23, opened 2021-04-01
minor postgres fix

fivetran-joemarkiewicz:
This PR includes the following update:

- Adjustment of the array_agg macro. It was originally thought that similar to redshift, postgres did not contain an array_agg function. However, it does and therefore the macro has been changed for the default to be array_agg and the exception to be redshift.


fivetran-jamie (approved):
lgtm!

PR #26, opened 2021-05-10
json parse addition

fivetran-joemarkiewicz:
 
This change introduces the `json_parse` macro. This macro is an overhauled version of the `json_extract` macro. The overhaul includes allowing the macro to take a list of items as the path to the value you want to abstract out of the json object. This should **NOT** simply be the `json_extract` macro as the macro will fail if a list in the form of brackets is passed through. As such, all existing `json_extract` macro instances will fail if that is the case.

I suggest we slowly replace our `json_extract` macros with this one as it is more dynamic and allows for more complex json extracting.

This macro was testing within the [dbt_shopify](https://github.com/fivetran/dbt_shopify/tree/transaction-order-updates) transaction-orders-updates branch package integration tests

This does not add a modification to an existing macro; however, I feel it should replace `json_extract` in the future.

- [x] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,19 @@ The data is returned by the path you provide as the argument. The json_extract m
 * `string_path`  (required): Name of the path in the json object which you want to extract the data from.
 
 ----
+### json_parse ([source](macros/json_parse.sql))
+This macro allows for cross database use of the json extract function, specifically used to parse and extract a nested value from a json object.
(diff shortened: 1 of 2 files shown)


fivetran-joemarkiewicz:
FYI the format decision to not indent is intentional. I tried many iterations of this macro with the `{%-` added to the for loop and conditionals, but still ran into compile errors unless I formatted the macro as you will see in the `json_parse.sql` file.

fivetran-jamie (approved):
lgtm

fivetran-kristin (approved):
LGTM

PR #28, opened 2021-05-24
Freshness tests

fivetran-jamie:
 
A `collect_freshness` macro to override [dbt's default freshness test](https://github.com/fishtown-analytics/dbt/blob/0.19.latest/core/dbt/include/global_project/macros/adapters/common.sql#L257-L273). This will allow us to incorporate any source-disabling variables so that `dbt source snapshot-freshness` does not fail due to these tables not existing. To disable the source `contact` table in hubspot, we'd add the following sort of `is_enabled` config to the packages source.yml
```yml
sources:
  - name: hubspot
    freshness:
      warn_after: {count: 84, period: hour}
      error_after: {count: 168, period: hour}
    tables:
      - name: contact
        meta:
          is_enabled: "{{ var('hubspot_marketing_enabled', true) and var('hubspot_contact_enabled', true) }}"
```

Addresses issues like https://github.com/fivetran/dbt_zendesk_source/issues/10 and https://github.com/fivetran/dbt_hubspot/issues/41. We can't use the solution put forth there, as the `sources` config block in a project's `dbt_project.yml` file would not capture the value of any referenced variable. It would always just automatically use the default value for the variable, since dbt thinks the variable = None at this point in execution. so it was basically hard-coding the source `+enabled` property to false and was not dynamically referring to any variable.

I tested it on Hubspot contact [data](https://github.com/fivetran/dbt_hubspot_source/blob/master/integration_tests/data/contact_data.csv), from a root dbt project that installed `hubspot_source` as a package. 

I ran `dbt source snapshot-freshness --select hubspot.contact` after setting the `hubspot_marketing_enabled` and/or `hubspot_contact_enabled` variables to false in my root `dbt_project.yml`. I did so after completely dropping the source table from my hubspot_schema as well, which did not cause freshness test issues when `hubspot_marketing_enabled` or `hubspot_contact_enabled` were set to false 

I also tested this out on the `fivetran_log` package -- sometimes `trigger_table` and/or `transformation` are missing if customers don't use fivetran transformations. I did the same thing I did with hubspot -- ran freshness tests before and after setting `fivetran_log_using_triggers` to false.

- [X] Yes 

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ than, or equal to, the specified numeric expression. The ceiling macro is compat
 
 ----
 ### collect_freshness ([source](macros/collect_freshness.sql))
-This macro overrides dbt's default [`collect_freshness` macro]((https://github.com/fishtown-analytics/dbt/blob/0.19.latest/core/dbt/include/global_project/macros/adapters/common.sql#L257-L273)) that is called when running `dbt source snapshot-freshness`. It allows you to incorporate model enabling/disabling variables into freshness tests, so that, if a source table does not exist, dbt will not run (and error on) a freshness test on the table. **Any package that has a dependency on fivetran_utils will use this version of the macro. If no `meta.is_enabled` field is provided, the `collect_freshness` should run exactly like dbt's default version.**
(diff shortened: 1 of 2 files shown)


fivetran-joemarkiewicz:
@fivetran-jamie this is awesome! Were you able to test this on a dataset where the downstream models are not disabled but the data is not fresh. Just to make sure the freshness test works as intended if the model is not disabled as this will be applied to all our packages once we merge to master.

fivetran-jamie:
@fivetran-joemarkiewicz yep! did so with hubspot seeded data and it worked as intended

and yeah i will re-run after merging with master -- it worked with the packages that kristin and i tested on (using this specific branch), but i think we're messing with dbt on a level where it'd be good to check that removing the specific `revision` from the `packages.yml` doesn't do anything unpredictable

fivetran-kristin (approved):

PR #30, opened 2021-06-16
Create union_data.sql

DylanBaker:


fivetran-joemarkiewicz (approved):
Thanks @DylanBaker just tested on Shopify and it worked using these macros!

When you get a chance would you be able to document within the README these two macros? In the meantime I will merge this PR so we can move forward with the Shopify PRs.

PR #37, opened 2021-07-13
v0.20 compatibility

### LINKED ISSUE #29: Changes to dispatch in dbt v0.20
Hi team! I wanted to give you a heads up about a change coming in dbt v0.20. Let me know if you have questions, I'm more than happy to help.

### Required code change

We've made a change to `adapter.dispatch`: Instead of passing a `_get_namespaces()` macro to the `packages` arg, you should pass the name of this package (as a string) to the `macro_namespace` arg, which is now the second positional argument.

In practice, this just looks like switching each macro like:
https://github.com/fivetran/dbt_fivetran_utils/blob/854b0f1d0b41d283fec7cdc486e8dfc86ddc14b5/macros/ceiling.sql#L1-L5

To:
```sql
{% macro ceiling(num) -%}

{{ adapter.dispatch('ceiling', 'fivetran_utils') (num) }}

{%- endmacro %} 
```

I hope this could be as simple as Find + Replace for `packages = fivetran_utils._get_utils_namespaces()` &rarr; `'fivetran_utils'`.

If you prefer more explicit syntax, you could specify keyword names for each argument. What's below is exactly the same as what's above:
```sql
{% macro ceiling(num) -%}

{{ adapter.dispatch(macro_name = 'ceiling', macro_namespace = 'fivetran_utils') (num) }}

{%- endmacro %} 
```

### For the README

If a user wishes to override/shim this package, instead of defining a var named `fivetran_utils_dispatch_list`, they should now define [a config](https://next.docs.getdbt.com/reference/project-configs/dispatch-config) in `dbt_project.yml`.

There's one crucial caveat: Your `_get_utils_namespaces` macro works just a little differently from everyone else's today. In order to maintain the current behavior of searching in `dbt_utils` _and then_ `fivetran_utils`, all users of Fivetran packages will need to specify at least this much:
```yml
dispatch:
  - macro_namespace: fivetran_utils
    search_order: ['dbt_utils', 'fivetran_utils']
```

They could add other packages to that list, too, for instance `spark_utils` or `<name_of_their_project>` (if they wish to override/reimplement certain macros).

One thought: It doesn't look like the `fivetran_utils` package currently requires `dbt_utils` in its `packages.yml` today. It really should, if you want to maintain the behavior of dispatching to `dbt_utils` before dispatching to `fivetran_utils`. If the `dbt_utils` package isn't installed, that will raise an error.

### Notes

This change is in dbt v0.19.2 as well. **Both v0.19.2 and v0.20.0 have backwards compatibility for the old syntax**, so there's no huge rush to upgrade. The old syntax will start raising a deprecation warning in v0.20. As soon as you do upgrade to the new syntax, you'll need to require `dbt >= 0.19.2` (or just `>=0.20.0`, for simplicity, since you're already making compatibility changes in #69).

See also:
- https://github.com/fishtown-analytics/dbt/issues/3383
- https://next.docs.getdbt.com/reference/dbt-jinja-functions/dispatch

jtcohen6:
- **Note:** This should actually go into a `releases/v0.2.latest` branch
- resolves #29

- Upgrade to new `adapter.dispatch` syntax
- Require `dbt>=0.20.0` and `dbt-utils: [">=0.7.0", "<0.8.0"]`
- Upgrade `version: 0.2.0` (cosmetic only)

This PR removes `_get_utils_namespaces`. It's no longer needed for the new `dispatch` syntax.

The way that this macro would prefer `dbt_utils` to `fivetran_utils` in the dispatch order... doesn't make a lot of sense to me! What was the purpose of this? In any case, it's possible for users to replicate this behavior by including the new `dispatch` syntax in their root `dbt_project.yml`:
```yml
dispatch:
  - macro_namespace: fivetran_utils
    search_order: ['dbt_utils', 'fivetran_utils']
```

- [x] Yes


fivetran-joemarkiewicz (approved):
Thanks so much for creating this PR and helping us on our efforts to make our packages dbt v0.20.0 compatible! 

Your changes look good and make sense to me. I just tried it out on my newly installed v0.20.0 dbt environment and worked like a charm. 🏅

PR #41, opened 2021-08-31
persist passthru macro

fivetran-jamie:
 
persist pass through column macro

in pendo

- [x] Yes

### DIFF
--- a/macros/persist_pass_through_columns.sql
+++ b/macros/persist_pass_through_columns.sql
@@ -1,0 +1,9 @@
+{% macro persist_pass_through_columns(pass_through_variable) %}
+
+{% if var(pass_through_variable) %}
+    {% for field in var(pass_through_variable) %}
+        , {{ field.alias if field.alias else field.name }}
+    {% endfor %}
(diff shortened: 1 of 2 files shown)


macros/persist_pass_through_columns.sql
```diff
@@ -0,0 +1,9 @@
+{% macro persist_pass_through_columns(pass_through_variable) %}
+
+{% if var(pass_through_variable) %}
```
fivetran-joemarkiewicz, line 3: should we add a default here in case the variable flow get's lost? My thought is below:
```suggestion
{% if var(pass_through_variable, none) %}
```
This way there is no chance dbt could throw the "variable does not exist" error.

fivetran-joemarkiewicz: Thoughts?

fivetran-joemarkiewicz (approved):
LGTM - Just one comment I want to get your thoughts on before we merge.

Also I seriously love this macro! I can't believe we hadn't done this before 🧠

PR #42, opened 2021-09-16
fixing add_pass_through_columns macro

jamesrayoub:
For instances where the source table does not include the additional pass through columns indicated in the project.yml file, the `fill_staging_columns()` macro will fail, specifically line 6-12: 
```
{% if column.name|lower in source_column_names -%}
    {{ fivetran_utils.quote_column(column) }} as 
    {%- if 'alias' in column %} {{ column.alias }} {% else %} {{ fivetran_utils.quote_column(column) }} {%- endif -%}
{%- else -%}
    cast(null as {{ column.datatype }})
    {%- if 'alias' in column %} as {{ column.alias }} {% else %} as {{ fivetran_utils.quote_column(column) }} {% endif -%}
{%- endif -%} 
```

This is because the pass through columns do not specify datatype and will result in compiled sql like `cast(null as )` which will fail. Modification of the `add_pass_through_columns` macro will allow for the addition of a datatype key in the project.yml  or use string as the default.

* tested in dbt_hubspot_source

- [ ] Yes
- [x] No (provide further explanation)

### DIFF
--- a/macros/add_pass_through_columns.sql
+++ b/macros/add_pass_through_columns.sql
@@ -4,7 +4,15 @@
 
     {% for column in pass_through_var %}
 
-      {% do base_columns.append({ "name": column.name, "alias": column.alias }) if column.alias else base_columns.append({ "name": column.name }) %}
+      {% if column.alias %}
+
+      {% do base_columns.append({ "name": column.name, "alias": column.alias, "datatype": column.datatype if column.datatype else dbt_utils.type_string()}) %}
+
+      {% else %}
+
+      {% do base_columns.append({ "name": column.name, "datatype": column.datatype if column.datatype else dbt_utils.type_string()}) %}
+        
+      {% endif %}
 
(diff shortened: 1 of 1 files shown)


jamesrayoub:
@fivetran-joemarkiewicz here is my suggested edit

jamesrayoub:
The reason for this PR is due to the fact that I'm trying to replicate the production models in a staging environment using staging instances of our tools which don't always have all of the same custom fields as in production; i.e. not every new custom field was tested in staging before creating in prod.

fivetran-joemarkiewicz (approved):
Thanks for opening this @jamesrayoub and the reasoning you provided makes sense.

I tested this and it looks good on my end! I do need to make a few minor edits before and merge to the appropriate branch prior to release. Therefore, I will merge to a staging feature branch and will merge to the release branch after. I will tag you in the following PR so you can be up to date.

PR #44, opened 2021-09-24
Feature/add passthrough update

fivetran-joemarkiewicz:
 
This PR adds the functionality to the `add_passthrough_columns` macro to allow for fields that do not exist to be included and set as a datatype. If the datatype is not set then it will default to string.

@jamesrayoub (the original creator of the PR #42) detailed the reason for this update:

> The reason for this PR is due to the fact that I'm trying to replicate the production models in a staging environment using staging instances of our tools which don't always have all of the same custom fields as in production; i.e. not every new custom field was tested in staging before creating in prod.

This was tested using the dbt_hubspot_source package.

- hubpost
- klaviyo
- pendo

- [ ] Yes

Edits to this README are not needed.

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -1,5 +1,5 @@
 name: 'fivetran_utils'
-version: '0.2.5'
+version: '0.2.6'
 config-version: 2
 
 source-paths: ["models"]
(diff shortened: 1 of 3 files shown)


fivetran-jamie (approved):
lgtm!

PR #46, opened 2021-10-01
Add ability to pass an array as a string to the union_data macro

DylanBaker:
 
Users have requested the ability to [dynamically generate schema lists for the `union_data` macro](https://github.com/fivetran/dbt_shopify/issues/25). Currently, this doesn't work because the macro gets passed the array of schemas as a string instead of as an array. 

This PR adds functionality to the macro so that we can parse the string and use it in the macro.

Tested it out within the Shopify package on my local machine.

- [ ] Yes
- [X] No, changes not necessary.

### DIFF
--- a/macros/union_data.sql
+++ b/macros/union_data.sql
@@ -9,6 +9,13 @@
 {% if var('union_schemas', none) %}
 
     {% set relations = [] %}
+    
+    {% if var('union_schemas') is string %}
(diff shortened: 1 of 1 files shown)


fivetran-joemarkiewicz:
Opened PR #47 to update the package version and be built off the `releases/v0.2.latest` branch.

PR #49, opened 2021-11-02
Expand union_data to allow multiple packges to leverage it with different variables

DylanBaker:
 
Currently a user couldn't use multiple packages simultaneously that use the union_data macro because they all look for the same variables (union_schemas and union_databases) which are hard-coded in the macro. This change makes it possible to dynamically set those variable names.  

Testing with downstream packages that I am working on.

It is being used in a number of existing packages. I have written the change such that it is backwards compatible.

- [ ] Yes


fivetran-joemarkiewicz (approved):
This looks great! @fivetran-jamie could you test this branch on your multi-database Klaviyo project to ensure the legacy union database feature works as well?

Since this PR was cut from `master` we will need to add the below package dependency in the source package.
```yml
packages:
  - git: "https://github.com/fivetran/dbt_fivetran_utils"
    revision: "feature/union-data-expansion"  
  - package: dbt-labs/dbt_utils
    version: 0.7.3
```

fivetran-jamie:
yep works for me with `union_databases`!

PR #54, opened 2021-12-06
Changes for dbt v1.0.0

DylanBaker:
 
Update package to be compatible with dbt v1.0.0. It bumps the dbt-utils version and makes a number of other small changes.

**N.B.:** I am currently merging this into `releases/v0.2.latest` which is probably incorrect. We should probably create a new `releases/v0.3.latest` for the new update.


fivetran-joemarkiewicz (approved):
Looks great to me, thanks for putting this PR together! 🙌

PR #56, opened 2022-01-19
collect freshness fix

fivetran-joemarkiewicz:
 
Fivetran created PR

This PR adjusts the `collect_freshness` macro to now refer to the `identifier` instead of the `name` of the node in the src.yml. This ensures if a dbt user has the `identifier` field populated, then the macro will work as intended. This has a high impact since this macro affects non Fivetran dbt packages. We will want to ensure this approach is a solid fix before merging. 

Additionally, we will most likely want to take a different approach to disabling freshness tests in our packages. The current solution is a too invasive and impacts parts of dbt projects that are outside the scope of Fivetran packages. We should use [this](https://github.com/dbt-labs/dbt-core/issues/4525) issue as inspiration for determining alternative approaches to freshness tests.

All source freshness tests that exist in users dbt projects if they have a Fivetran dbt package installed.

- [ ] Yes

This is an under the hood update and does not need a README update.

### DIFF
--- a/macros/collect_freshness.sql
+++ b/macros/collect_freshness.sql
@@ -19,9 +19,8 @@
     select
       {% if is_enabled %}
       max({{ loaded_at_field }})
-      {{ current_timestamp() }}
       {% else %} 
-      max({{ loaded_at_field }}) {% endif %} as max_loaded_at,
+      {{ current_timestamp() }} {% endif %} as max_loaded_at,
       {{ current_timestamp() }} as snapshotted_at
 
     {% if is_enabled %}
--- a/macros/collect_freshness.sql
+++ b/macros/collect_freshness.sql
@@ -8,7 +8,7 @@
 
   {%- set enabled_array = [] -%}
   {% for node in graph.sources.values() %}
-    {% if node.name == source.name %}
+    {% if node.identifier == source.identifier %}
       {% if (node.meta['is_enabled'] | default(true)) %}
         {%- do enabled_array.append(1) -%}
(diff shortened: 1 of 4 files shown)


macros/collect_freshness.sql
```diff
@@ -8,7 +8,7 @@
 
   {%- set enabled_array = [] -%}
   {% for node in graph.sources.values() %}
-    {% if node.name == source.name %}
+    {% if node.identifier == source.identifier %}
```
fivetran-jamie, line 11: if an identifier or alias isn't explicitly provided to a source, is `identifier` the `name` by default? @fivetran-joemarkiewicz

fivetran-joemarkiewicz: That is what I got from my testing. When using this on nodes with **only** the `name` config, it just defaulted to use the name.

fivetran-jamie: ah nice seems like it does

macros/collect_freshness.sql
```diff
@@ -17,10 +17,10 @@
   {% set is_enabled = (enabled_array != []) %}
 
     select
-      {% if is_enabled %}
-      max({{ loaded_at_field }})
+      {% if is_enabled == false %}
```
fivetran-jamie, line 20: why this change? just wanna make sure there's not an issue with the conditional on line 26 as well

fivetran-jamie (approved):

fivetran-jamie (approved):
looks good!

PR #58, opened 2022-01-21
Fixed: union_data macro

lizdeika:
Fixes `union_data` macro

- [ ] Yes

### DIFF
--- a/macros/union_data.sql
+++ b/macros/union_data.sql
@@ -6,11 +6,11 @@
 
 {% macro default__union_data(table_identifier, database_variable, schema_variable, default_database, default_schema, default_variable) %}
 
-{% if var('union_schemas', none) %}
+{% if var('shopify_union_schemas', none) %}
 
     {% set relations = [] %}
 
-    {% for schema in var('union_schemas') %}
+    {% for schema in var('shopify_union_schemas') %}
 
     {% set relation=adapter.get_relation(
         database=var(database_variable, default_database),
@@ -30,11 +30,11 @@
 
     {{ dbt_utils.union_relations(relations) }}
 
-{% elif var('union_databases', none) %}
+{% elif var('shopify_union_databases', none) %}
 
     {% set relations = [] %}
 
-    {% for database in var('union_databases') %}
+    {% for database in var('shopify_union_databases') %}
 
     {% set relation=adapter.get_relation(
         database=database,


fivetran-joemarkiewicz:
Hi @lizdeika thanks for opening this PR!

I see the edit you have made is to the name of the variable itself. Are you seeing the jobs are failing on your end with this variable being named differently? We have designed this macro to be dynamically different for each package it is being used in. I want to make sure that is the case and there isn't an unintended bug.

fivetran-joemarkiewicz:
Additionally, the branch that is actually used in the releases of this package is now `releases/v0.3.latest` I should probably add this as a disclaimer in the README 😄

lizdeika:
Hi,
`union_schemas` was renamed to `shopify_union_schemas` in `dbt_shopify` package https://github.com/fivetran/dbt_shopify/blob/main/CHANGELOG.md
So this method can not find the `union_schemas` when dbt_shopify is upgraded and goes to the last `else`: https://github.com/fivetran/dbt_fivetran_utils/blob/eb4c239d71ea3f56f61974a284c4b6d8ae973471/macros/union_data.sql#L57

fivetran-joemarkiewicz:
@lizdeika in updating to [Shopify v0.6.0](https://github.com/fivetran/dbt_shopify/releases/tag/v0.6.0) you actually just need to update the variable name in your `dbt_project.yml` to be `shopify_union_schemas` now! You can see an example of what to update [here](https://github.com/fivetran/dbt_shopify#configuration)
```txt
# dbt_project.yml

...
config-version: 2

vars:
    shopify_union_schemas: ['shopify_usa','shopify_canada'] same database/project
    shopify_union_databasesv: ['shopify_usa','shopify_canada']
```

This change was made in order to scale this solution. As we add the union data ability to other packages we wanted the variable names to be unique. Therefore, we won't need to update the macro in this package. Let me know if you have any questions!

lizdeika:
@fivetran-joemarkiewicz updating to `shopify_union_schemas` is exactly the reason why this package stops working.
I am now forced to use both `shopify_union_schemas`(for dbt shopify 0.6.0) and a duplicate `union_schemas` (for dbt fivetran utils 0.8.0) with the same values to work them together.

```
vars:
  union_schemas: ['schema1', 'schema2', 'schema3']
  shopify_union_schemas: ['schema1', 'schema2', 'schema3']
```

fivetran-joemarkiewicz:
Hmm 🤔 That shouldn't be needed.

Looking at your above questions I actually noticed that you are mentioning the `fulfillment_source` tmp model. However, we do not have this model in our base package. Is this by chance a model you added within a fork? Additionally, when we made the changes to the `union_data` macro, we had to apply updates to all our tmp models to account for this change. Looking at the tmp model code you shared, it doesn't seem to have the reflected updates.

https://github.com/fivetran/dbt_shopify_source/blob/c7b95b8a7d832e4a3cdbe2867dc6c52561b578e5/models/tmp/stg_shopify__order_tmp.sql#L9-L10

If the `fulfillment_source` model is part of a fork, then you will want to add the `union_schema_variable` and `union_database_variable` arguments to your tmp model.

lizdeika:
I see! `union_schema_variable` did not exist when we made our custom _tmp model.
Thanks! This solves the problem.

PR #64, opened 2022-02-23
automating docs generate

fivetran-sheringuyen:
 
Adds new macro to automate docs generate part of package workflow

Copied an existing package and ran the following tests from adjacent development directory and installed local copy of dbt fivetran utils:
1. Updating docs: Replaces existing project/docs contents with new docs
2. Creating docs: Creates new docs directory and moves integration_tests/target/ docs files to new docs folder

New macro

- [x] Yes


README.md
```diff
@@ -578,6 +600,21 @@ source dbt_packages/fivetran_utils/generate_columns.sh '../dbt_apple_search_ads_
 In that example, it will:
 * Create a `get_campaign_history_columns.sql` file in the `macros` directory, with the necessary macro within it.
 
+----
+
+### generate_docs.sh([source](generate_docs.sh))
+
+This bash file can be used to create or replace package documentation (`<project_name>/docs`).
```
fivetran-joemarkiewicz, line 607: It isn't clear where I need to run this command. Do I run it in the direction of the `integration_tests` folder, or within the root of my `development` dbt project? Can we specify that here.

generate_docs.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+ORIG_DIR=$(echo $PWD)
+PKG=$(echo $1)
+cd ${PKG}/integration_tests
+dbt clean && dbt deps && dbt seed && dbt run
```
fivetran-joemarkiewicz, line 5: Do we need to run `dbt run`? Could we limit the steps by just executing `dbt clean && dbt deps && dbt seed && dbt docs generate`?

fivetran-joemarkiewicz (changes requested):
This is super cool @fivetran-sheringuyen!

I am running into a bit of a blocker on getting it to work however. Please see my inline comments. Additionally, below is the result I get when running the source command.
image

The end result then removes the files from my `/docs` folder, but then doesn't regenerate them?

README.md
fivetran-sheringuyen: Sure thing! Where is your development folder located wrt the dbt_jira_source directory?

fivetran-sheringuyen: image
This macro is designed to work in all directories under `stripe` (the project directory) that exist at the same level as `dbt_stripe`. So you can run this from `dev` which exists adjacent to `dbt_stripe`. I hope that helps! If it still doesn't work let me know. 

@fivetran-joemarkiewicz

fivetran-joemarkiewicz (approved):
Looks great! 

One thing to add before merging is highlighting the fact that we need to update our target schema in `profiles.yml`

PR #72, opened 2022-04-04
Releases/v0.3.6

fivetran-reneeli:
 
Further automating the staging model creation to prefill column fields in the final select statement.

To test new macro `get_column_names_only`, use
`dbt run-operation get_column_names_only --args '{table_name: log, schema_name: fivetran_log, database_name: dbt-package-testing}' | tail -n +2 > fivetran_log/models/log__log_temp.sql`

Then to test how the actual bash script `generate_models` gets updated, use
`source dbt_packages/fivetran_utils/generate_models.sh '../dbt_fivetran_log_source' stg_fivetran_log dbt-package-testing fivetran_log log`

Tested using fivetran log

- [x] Yes

Updated the readme with adding the new macro + updated affected macros

### DIFF
--- a/README.md
+++ b/README.md
@@ -527,7 +527,7 @@ Note this will retain the timestamp from the built-in formatting update from dbt
 
 **Usage:**
 ```bash
-dbt run-operation get_column_names_only --args '{table_name: [table_name], schema_name: [schema_name], database_name: [database_name]}' | tail -n +2 
+dbt run-operation get_column_names_only --args '{table_name: log, schema_name: fivetran_log, database_name: database-name' 
 
 ```
 **CLI Output:**
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ dispatch:
 
 - [Automations](#automations)
   - [generate_columns_macro](#generate_columns_macro-source)
+  - [get_column_names_only](#get_column_names_only-source)
   - [get_columns_for_macro](#get_columns_for_macro-source)
   - [generate_docs](#generate_docs-source)
   - [staging_models_automation](#staging_models_automation-source)
--- a/README.md
+++ b/README.md
@@ -519,6 +519,37 @@ source dbt_packages/fivetran_utils/generate_docs.sh '../dbt_apple_search_ads_sou
 * `package` (required): Name of the package; include whether package is source or not
 
 ----
+### get_column_names_only ([source](macros/get_column_names_only.sql))
+This macro is used in the `generate_models.sh` script to further the `staging_models_automation` macro. This macro outputs all columns from the specified table, allowing `generate_models.sh` to prefill column fields in the final select statement.
+
+Note this will retain the timestamp from the built-in formatting update from dbt 1.0.0. Therefore in the staging model resulting from `generate_models.sh`, you will need to manually delete the timestamp.
+
+**Usage:**
+```bash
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -518,6 +519,37 @@ source dbt_packages/fivetran_utils/generate_docs.sh '../dbt_apple_search_ads_sou
...
 
+----
+### get_column_names_only ([source](macros/get_column_names_only.sql))
+This macro is used in the `generate_models.sh` script to further the `staging_models_automation` macro. This macro outputs all columns from the specified table, allowing `generate_models.sh` to prefill column fields in the final select statement.
+
+Note this will retain the timestamp from the built-in formatting update from dbt 1.0.0. Therefore in the staging model resulting from `generate_models.sh`, you will need to manually delete the timestamp.
+
+**Usage:**
+```bash
+dbt run-operation get_column_names_only --args '{table_name: [table_name], schema_name: [schema_name], database_name: [database_name]}' | tail -n +2
```
fivetran-joemarkiewicz, line 530: When reading this I would imagine I need to add brackets when inserting my arguments. So I would think I need to do:
```bash
dbt run-operation get_column_names_only --args '{table_name: [log], schema_name: [fivetran_log], database_name: [my-database]}'
```
However, I think we need to use single quotes instead. Can you update this to be similar to a real world example of how to use this.

fivetran-joemarkiewicz: Also, is the `tail -n +2` needed in the usage?

generate_models.sh
```diff
@@ -18,15 +18,20 @@ fields as (
...
+
+dbt run-operation fivetran_utils.get_column_names_only --args '{"table_name": "'$5'", "schema_name": "'$4'", "database_name":"'$3'"}' | tail -n +2 >> $1/models/$2__$5.sql
+    
+    
+echo "    from fields
 )
 
-select * from final" >> $1/models/$2__$5.sql
+select *
+from final" >> $1/models/$2__$5.sql
```
fivetran-joemarkiewicz, line 36: Is this in line with our current sql style standards?

fivetran-joemarkiewicz (approved):
Great job @fivetran-reneeli 🙌 

This PR looks great and I think this will really help with our automation efforts for the source package build. Say hello to quicker and more efficient build times 😄 !

Before we merge and release this, please address a few of the comments I have below. Also, it would be great if you could showcase this during tomorrows standup if there is time. Especially since I am unable to test the macro with my computer dying 😭

generate_models.sh
fivetran-reneeli: @fivetran-joemarkiewicz so i'm basing it off the sql guide, using this as reference: https://github.com/fivetran/dbt_style_guide/blob/main/sql_guidelines.md#ctes

README.md
fivetran-reneeli: Updated to use real example!

Also, so this is what it looks like without that formatting at the end. But you're right in that we don't actually need to include the formatting bit in the macro description itself, so I'll remove it here.
```
21:19:03  Running with dbt=1.0.0
21:19:03  Unable to do partial parsing because a project config has changed
21:19:05          _fivetran_synced,
        connector_id,
        event,
        id,
        message_data,
        message_event,
        process_id,
        sequence_number,
        sync_id,
        time_stamp,
        transformation_id

```

PR #75, opened 2022-04-14
expand try_cast for snowflake

fivetran-jamie:
 

for snowflake, the native try_cast function can only convert string-type fields to another datatype (limited to the ones [here](https://docs.snowflake.com/en/sql-reference/functions/try_cast.html#usage-notes)). so if we try to convert a float to a numeric, such as in this issue https://github.com/fivetran/dbt_klaviyo/issues/21 snowflake throws an error.

my idea is to cast the field-arg as a string proactively. curious if there are concerns around losing precision with doing so though.....

- [ ] Yes

### DIFF
--- a/macros/try_cast.sql
+++ b/macros/try_cast.sql
@@ -40,7 +40,7 @@
 {% endmacro %}
 
 {% macro snowflake__try_cast(field, type) %}
-    try_cast({{field}} as {{type}})
+    try_cast(cast({{field}} as varchar) as {{type}})
 {% endmacro %}
 
 {% macro bigquery__try_cast(field, type) %}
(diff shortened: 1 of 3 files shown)


fivetran-jamie:
wait to merge until reinvestigating the issues from the 0.3.6 release

macros/try_cast.sql
```diff
@@ -40,7 +40,7 @@
 {% endmacro %}
 
 {% macro snowflake__try_cast(field, type) %}
-    try_cast({{field}} as {{type}})
+    try_cast(cast({{field}} as varchar) as {{type}})
```
fivetran-jamie, line 43: alternatively

```suggestion
    case 
        when left(system$typeof({{field}}), 7) = 'VARCHAR' then try_cast({{field}} as {{type}})
        else cast({{field}} as {{type}}) end
```

fivetran-jamie: the original code works for the customer who raised the issue btw!

fivetran-joemarkiewicz:
This was integrated within the `v0.3.8` release of the package. Closing the Issue.

PR #81, opened 2022-07-28
Feature/transform persist

fivetran-jamie:
 
added a `transform` argument to the transform package passthrough column macro, so that we could aggregate them

tested in Pendo and other packages 

pendo, twitter_ads 

- [x] Yes


macros/persist_pass_through_columns.sql
```diff
@@ -1,8 +1,8 @@
-{% macro persist_pass_through_columns(pass_through_variable) %}
+{% macro persist_pass_through_columns(pass_through_variable, identifier, transform='') %}
```
fivetran-joemarkiewicz, line 1: This works, but I would feel more comfortable if we added a default value for `identifier`, just to be safe.

fivetran-joemarkiewicz (approved):
Just left the one comment to add a default value for `identifier`. Once that is applied we will be good to go!

I do want to hold off merging until Monday just so we are prepared in case anything comes up.

macros/persist_pass_through_columns.sql
fivetran-jamie: added identifier=none

fivetran-joemarkiewicz (changes requested):
@fivetran-jamie I have tested this some more and it looks good to go!

Before we cut this release would you mind updating the following:
- [x] Remove the .DS_Store file from the PR.
- [x] Add a CHANGELOG entry.

fivetran-jamie:
@fivetran-joemarkiewicz updates are made!

fivetran-joemarkiewicz (approved):
LGTM

Issue #52, opened 2021-11-18
BUG - Freshness tests fail when `union_schemas` or `union_databases` declared in Fivetran plugins


pawelngei:
**Are you a current Fivetran customer?** 
Pawel Ngei, Pattern Brands

**Describe the bug**
If you're using one of the Fivetran DBT plugins which supports multiple schemas / databases, like Shopify, Klaviyo, (soon) Google Ads or Facebook Ads, the freshness tests defined in `_source` package fail as there is no `{{ var ('klaviyo_schema', 'klaviyo') }}` defined.

**Steps to reproduce**
1. Set up a project with multiple schemas
2. Use one of the `_source` plugins from Fivetran, like `dbt_klaviyo_source`.
3. Provide schemas via `union_schemas` in your `dbt_project.yml`, with no single `klaviyo_schema` and no schema called just `klaviyo`. Let's make them just `klaviyo_1` and `klaviyo_2`.
4. Run `dbt source snapshot-freshness` or` dbt source freshness`.
5. The Klaviyo plugin freshness test will fail

**Expected behavior**
The Klaviyo freshness plugin will test the freshness of multiple schemas about to be unioned.

**Project variables configuration**

```yml
vars:
  klaviyo_source:
    klaviyo_database: 'RAW'
    union_schemas: [
      'KLAVIYO_1',
      'KLAVIYO_2',
      'KLAVIYO_3',
      'KLAVIYO_4'
    ]
```

**Package Version**
```yml
packages:
  - package: fivetran/klaviyo_source
    version: 0.3.0
```

**Warehouse**
- [ ] BigQuery
- [ ] Redshift
- [x] Snowflake
- [ ] Postgres
- [ ] Databricks
- [ ] Other (provide details below)

**Additional context**
I believe that this bug can be solved using this utils repository, will describe it in a comment below.

**Please indicate the level of urgency** 
Will require this feature before we can call our setup "production-ready", so effectively just after Google Ads and Facebook Ads unioning PRs are merged.

**Are you interested in contributing to this package?** 
- [ ] Yes, I can do this and open a PR for your review.
- [x] Possibly, but I'm not quite sure how to do this. I'd be happy to do a live coding session with someone to get this fixed. 
- [ ] No, I'd prefer if someone else fixed this.  I don't have the time and/or don't know what the root cause of the problem is.

pawelngei:
[...]

PR #82, opened 2022-08-18
bugfix/union-data

fivetran-joemarkiewicz:
 
## Bug Fixes
- The `union_data` macro has been adjusted to establish a source relation instead of a floating relation. 
    - To leverage this macro in an environment where `dbt_fivetran_utils` is already installed, you can add the following to your `packages.yml` to install this version of the package:
    ```yml
    - git: https://github.com/fivetran/dbt_fivetran_utils.git
      revision: bugfix/union-data
      warn-unpinned: false
    ```
    - You can then add the following to your `dbt_project.yml` to dispatch this package version (`fivetran_utils_union`) in place of the original package (`dbt_fivetran_utils`).
    ```yml
    dispatch:
      - macro_namespace: fivetran_utils
        search_order: ['fivetran_utils_union','fivetran_utils']
    ```

This is not a new macro, instead it is changes to the existing `union_data` macro.

This was tested on the Xero dataset and success was found! The key update needed here however is a source.yml for each source **must** be created. See below for the two source.yml I created in my development project (notice the only changes are the source name and schema name):
image

In the end you can see the union working as intended!
image

- [ ] Yes

No README changes required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+# dbt_fivetran_utils v0.4.0-b1
+## Bug Fixes
+- The `union_data` macro has been adjusted to establish a source relation instead of a floating relation. 
+    - To leverage this macro in an environment where `dbt_fivetran_utils` is already installed, you can add the following to your `packages.yml` to install this version of the package:
+    ```yml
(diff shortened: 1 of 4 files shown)


fivetran-joemarkiewicz:
I have a strong feeling this PR would also address Issue #52

PR #89, opened 2022-09-29
Bugfix/union data using sources if defined

fivetran-jamie:
 
Allows for the `union_data` macro to establish relationships between sources and the models in which they are unioned. The default behavior is still to use a floating relation, but sources will be used if the customer sets `has_defined_sources` to `true` (scoped to the package in which the macro is being used) and has defined the sources.

i tested it on the Xero package, setting `has_defined_sources` to true/false and adding/removing source definitions on the `account` table

ensured that the default behavior remains the same.  For pacakges that have unioning capabilities, we'll need to add stuff about the new `has_defined_sources` variable to their READMEs

- [x] Yes


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# dbt_fivetran_utils v0.3.10
+
+## Bug Fixes
+- The `union_data` macro has been adjusted to establish a source relation instead of a floating relation, though the default behavior still uses a floating relation. To establish relationships between sources and the models that union them:
+  - Define the sources in `.yml` files in your project. Only the names of tables are required.
+  - In your `dbt_project.yml` file, set the `has_defined_sources` variable to `true`. This variable has a generic name, so you must scope it to the package/project in which the `union_data` macro is called.
```
fivetran-joemarkiewicz, line 7: Can you reference this PR in the CHANGELOG entires.

README.md
```diff
@@ -428,6 +428,8 @@ If the `var` with the name of the `database_variable` argument is set, the macro
 
 When using this functionality, every `_tmp` table should use this macro as described below.
 
+To create dependencies between the unioned model and its **sources**, you must define the source tables in a `.yml` file in your project and set the `has_defined_sources` variable (scoped to the source package in which the macro is being called) to `True` in your `dbt_project.yml` file.
```
fivetran-joemarkiewicz: Can you add a bit more documentation here for how this can be configured. Maybe an extra section that shows how this will be defined in the `dbt_project.yml` and a scenario of three sources being unioned with this new feature from an example package?

fivetran-joemarkiewicz: Can you also bold the `must define` to really make the impact that this will not work if the variable is set to true and the sources are not defined.

dbt_project.yml
```diff
@@ -1,4 +1,4 @@
 name: 'fivetran_utils'
-version: '0.3.9'
+version: '0.3.10'
```
fivetran-joemarkiewicz: We will want to fold this into `0.4.0` instead.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# dbt_fivetran_utils v0.3.10
```
fivetran-joemarkiewicz: Let's incorporate this in our next breaking change `v0.4.0`

fivetran-joemarkiewicz: I have updated the base branch for this PR accordingly.

fivetran-joemarkiewicz (changes requested):
Thanks for adding this functionality @fivetran-jamie! I have a few requests for documentation updates before this will be fully ready to go!

Overall I tested these changes and found it to work perfectly!! I just want to make sure we make it apparent in our documentation how you are able to leverage this variable.

fivetran-joemarkiewicz (approved):
@fivetran-jamie thanks so much for applying these updates! These updates look great and I appreciate your detail on the README section for how to effectively use the two versions of this macro.

Be sure to resolve the merge conflicts in the CHANGELOG prior to merging. Other than that, this is good to go!

PR #92, opened 2022-11-07
add find and replace to json extract

fivetran-reneeli:
 
Add a find+replace to our pivot_json_extract macro, similar to `(' ', '_')` but with `('.', '_')`.  

n/a

Stripe
Issue https://github.com/fivetran/dbt_stripe/issues/45
Testing with `dbt compile -t snowflake  -m stg_stripe__plan ` and `dbt run -t snowflake  -m stg_stripe__plan     `

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# dbt_fivetran_utils v0.4.2
+
+## Updates
+[PR #92](https://github.com/fivetran/dbt_fivetran_utils/pull/92) updates the `pivot_json_extract` macro to include additional formats of fields to be pivoted out into columns. Specifically, allowing for fields with `.` to be replaced with `_`, for metadata variables to accept dictionaries in addition to strings, and for aliasing of fields.
+
 # dbt_fivetran_utils v0.4.1
 
 ## Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,10 @@
-# dbt_fivetran_utils v0.4.2
-
-## Updates
-[PR #92](https://github.com/fivetran/dbt_fivetran_utils/pull/92) updates the `pivot_json_extract` macro to include additional formats of fields to be pivoted out into columns. Specifically, allowing for fields with `.` to be replaced with `_`, for metadata variables to accept dictionaries in addition to strings, and for aliasing of fields.
-
 # dbt_fivetran_utils v0.4.1
 
 ## Bug Fixes
 [PR #91](https://github.com/fivetran/dbt_fivetran_utils/pull/91) updates dispatch from `{{ dbt_utils.<macro> }}` to `{{ dbt.<macro> }}` for additional [cross-db macros](https://docs.google.com/spreadsheets/d/1xF_YwJ4adsnjFkUbUm8-EnEL1r_C-9_OI_pP4m4FlJU/edit#gid=1062533692) missed in the `fivetran_utils.union_relations()` macro.
+
+## Updates
+[PR #92](https://github.com/fivetran/dbt_fivetran_utils/pull/92) updates the `pivot_json_extract` macro to include additional formats of fields to be pivoted out into columns. Specifically, allowing for fields with `.` to be replaced with `_`, for metadata variables to accept dictionaries in addition to strings, and for aliasing of fields.
 
 # dbt_fivetran_utils v0.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 # dbt_fivetran_utils v0.4.1
 
 ## Bug Fixes
-[PR #91](https://github.com/fivetran/dbt_fivetran_utils/pull/91) updates dispatch from `{{ dbt_utils.<macro> }}` to `{{ dbt.<macro> }}` for additional [cross-db macros](https://docs.google.com/spreadsheets/d/1xF_YwJ4adsnjFkUbUm8-EnEL1r_C-9_OI_pP4m4FlJU/edit#gid=1062533692) missed in the `fivetran_utils.union_relations()` macro.
+- Updates dispatch from `{{ dbt_utils.<macro> }}` to `{{ dbt.<macro> }}` for additional [cross-db macros](https://docs.google.com/spreadsheets/d/1xF_YwJ4adsnjFkUbUm8-EnEL1r_C-9_OI_pP4m4FlJU/edit#gid=1062533692) missed in the `fivetran_utils.union_relations()` macro. ([PR #91](https://github.com/fivetran/dbt_fivetran_utils/pull/91))
 
 ## Updates
-[PR #92](https://github.com/fivetran/dbt_fivetran_utils/pull/92) updates the `pivot_json_extract` macro to include additional formats of fields to be pivoted out into columns. Specifically, allowing for fields with `.` to be replaced with `_`, for metadata variables to accept dictionaries in addition to strings, and for aliasing of fields.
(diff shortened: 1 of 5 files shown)


fivetran-joemarkiewicz (changes requested):
@fivetran-reneeli your approach here makes sense and definitely addresses the `feature.id` type metadata fields that the customer was experiencing. However, as you pointed out, it won't solve the reserved word error or other errors we may not be aware of at the moment.

I was able to do some digging and think I found a unique workaround that can address both the `.` issue and the reserved word error. See below for the modified macro code I added to make this work:

```sql
{% macro pivot_json_extract(string, list_of_properties) %}

{%- for property in list_of_properties -%}
{%- if property is not string -%}
replace( {{ fivetran_utils.json_extract(string, property.name) }}, '"', '') as {{ property.alias if property.alias else property.name | replace(' ', '_') | replace('.', '_') | lower }}

{%- else -%}
replace( {{ fivetran_utils.json_extract(string, property) }}, '"', '') as {{ property | replace(' ', '_') | lower }}

{%- endif -%}
{%- if not loop.last -%},{%- endif %}
{% endfor -%}

{% endmacro %}
```

With this update to the macro, we allow the variable to accept a dictionary in addition to the normal string behavior. This means, if a customer wants to alias the fields or is experiencing a reserved word or incompatible name, then they can change it to their liking!

I was able to get this to work using the following variable config:
```yml
  stripe__plan_metadata:
    - limit
    - version
    - name: feature.id
      alias: feature_id
    - name: plan.handle
      alias: plan_handle
    - name: set
      alias: sets
```
I feel we should update the macro for the next iteration of fivetran_utils to follow this behavior. This could then address the customers issue. We would just want to update the dbt_stripe and dbt_stripe_source README's to show how a customer may leverage these new dictionary arguments if needed.

fivetran-reneeli:
Hi @fivetran-joemarkiewicz , I updated the readmes in the stripe packages to reflect these new instructions. Let me know if they make sense! 
[stripe](https://github.com/fivetran/dbt_stripe/pull/51)
[stripe source](https://github.com/fivetran/dbt_stripe_source/pull/57)

fivetran-joemarkiewicz:
Thanks @fivetran-reneeli I noticed this PR didn't apply the updates to the macro. Would you be able to make those updates as well?

macros/pivot_json_extract.sql
```diff
@@ -1,9 +1,13 @@
 {% macro pivot_json_extract(string, list_of_properties) %}
 
 {%- for property in list_of_properties -%}
+{%- if property is not string -%}
```
fivetran-joemarkiewicz: I think we discussed during that quick show and tell that it may be better to use the following. This way you must be using a dictionary instead of not string. This way if customers need to pass an int or something, they aren't hit with an error.
```suggestion
{%- if property is mapping -%}
```

fivetran-joemarkiewicz (changes requested):
@fivetran-reneeli I had one comment that I included in the notes below. Can you test this on your end and make sure it works as expected.

Additionally, be sure to add a CHANGELOG entry. Let me know once the updates are applied and I can re-review.

macros/pivot_json_extract.sql
fivetran-reneeli: Oh yes that's right-- testing it now

fivetran-reneeli:
Test with the following packages.yml in dbt_stripe_source

```.yml
packages:
# - package: fivetran/fivetran_utils
#   version: [">=0.4.0", "<0.5.0"]
- package: dbt-labs/spark_utils
  version: [">=0.3.0", "<0.4.0"]

- git: https://github.com/fivetran/dbt_fivetran_utils.git
  revision: bugfix/json_replace
  warn-unpinned: false

**Comment by fivetran-reneeli** [comment]
And adding this to the project yml:

```.yml
    stripe__plan_metadata:
        - name: limit
          alias: limit_z
        - name: feature.id
          alias: feature_id
        - name: plan.handle
          alias: plan_handle
        - name: set
          alias: sets
        - name: 123
          alias: onetwothree

**Comment by fivetran-reneeli** [comment]
Hi @fivetran-joemarkiewicz , I made the updates and also updated the changelog and readmes and versioning. Lmk if it looks good!

File: CHANGELOG.md [hunk 2]
```diff
@@ -1,3 +1,8 @@
+# dbt_fivetran_utils v0.4.2
```
fivetran-joemarkiewicz: 0.4.2 hasn't been released yet. We should fold these into the 0.4.1 changelog entries instead.

dbt_project.yml
```diff
@@ -1,4 +1,4 @@
 name: 'fivetran_utils'
-version: '0.4.1'
+version: '0.4.2'
```
fivetran-joemarkiewicz: 0.4.1 isn't released yet, we can keep this as 0.4.1

integration_tests/dbt_project.yml
```diff
@@ -1,4 +1,4 @@
 name: 'fivetran_utils_integration_tests'
-version: '0.4.1'
+version: '0.4.2'
```
fivetran-joemarkiewicz: Same as above, we can keep as 0.4.1

fivetran-joemarkiewicz (changes requested):
@fivetran-reneeli just a few small comments regarding the versioning of the package and folding the changelog entries into the soon to be released 0.4.1 entry.

Let me know once the updates have been applied and I can re-review

CHANGELOG.md
fivetran-reneeli: updated

dbt_project.yml
fivetran-reneeli: updated

integration_tests/dbt_project.yml
fivetran-reneeli: updated

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## Bug Fixes
 [PR #91](https://github.com/fivetran/dbt_fivetran_utils/pull/91) updates dispatch from `{{ dbt_utils.<macro> }}` to `{{ dbt.<macro> }}` for additional [cross-db macros](https://docs.google.com/spreadsheets/d/1xF_YwJ4adsnjFkUbUm8-EnEL1r_C-9_OI_pP4m4FlJU/edit#gid=1062533692) missed in the `fivetran_utils.union_relations()` macro.
 
+## Updates
+[PR #92](https://github.com/fivetran/dbt_fivetran_utils/pull/92) updates the `pivot_json_extract` macro to include additional formats of fields to be pivoted out into columns. Specifically, allowing for fields with `.` to be replaced with `_`, for metadata variables to accept dictionaries in addition to strings, and for aliasing of fields.
```
fivetran-joemarkiewicz: Not sure why we went with the PR ref being at the start of the sentence here 🤔 

Would you be able to move both of these to the typical end of the sentence if it is just one change we are highlighting?
```suggestion
- Updates dispatch from `{{ dbt_utils.<macro> }}` to `{{ dbt.<macro> }}` for additional [cross-db macros](https://docs.google.com/spreadsheets/d/1xF_YwJ4adsnjFkUbUm8-EnEL1r_C-9_OI_pP4m4FlJU/edit#gid=1062533692) missed in the `fivetran_utils.union_relations()` macro. ([PR #91](https://github.com/fivetran/dbt_fivetran_utils/pull/91))

## Updates
- Updates the `pivot_json_extract` macro to include additional formats of fields using a new `name` and `alias` argument to be pivoted out into columns. Specifically, allowing for fields with `.` to be replaced with `_`, for metadata variables to accept dictionaries in addition to strings, and for aliasing of fields. ([PR #92](https://github.com/fivetran/dbt_fivetran_utils/pull/92))
```

fivetran-joemarkiewicz (approved):
I just have one small request to update the CHANGELOG. Once you commit that suggestion, this will be good to merge into `releases/v0.4.latest`

PR #93, opened 2022-12-13
Backwards compat passthru

fivetran-jamie:
 

Makes staging passthrough column macros backwards compatible for lists of passthrough columns (instead of dictionaries with the name and alias and transform)

a bunch but i tested on shopify

- [x] Yes


fivetran-reneeli (approved):
Looks good!

Issue #99, opened 2023-02-14
make union_data more robust 


fivetran-jamie:
see https://github.com/fivetran/dbt_shopify_source/issues/57

When no sources in a union_schema or union_database array have a table (or if the first schema/db in the array doesn't have it?) our models fail, because we are not doing the `cast(null as <data_type>) as field_in_table_we_dont_have` thing. I think we should add a parameter to the union_data macro that is a dictionary/json very similar to what we have in the `get_<table>_columns` macros. then union_data could handle the null stuff

PR #100, opened 2023-03-03
union data creates empty tables if none exist

fivetran-jamie:
 
- adds argument to union_data to handle when there's no table in any of the provided schemas/databases (by creating an empty table) 

https://github.com/fivetran/dbt_fivetran_utils/issues/99 (came out of https://github.com/fivetran/dbt_shopify_source/issues/57)

on shopify

Packages that use the `union_data` macro (found by searching `union_data` on github within the Fivetran org), and the impact of this change on them:
- Facebook pages
  - If a table doesn’t exist, in previous versions the package would have failed. Now it creates an empty ones and works with those
- Twitter organic  
  - If a table doesn’t exist, in previous versions the package would have failed. Now it creates an empty ones and works with those
- Instagram business
  - If a table doesn’t exist, in previous versions the package would have failed. Now it creates an empty ones and works with those
- Linkedin pages 
  - If a table doesn’t exist, in previous versions the package would have failed. Now it creates an empty ones and works with those
- Stripe
  - Instead of dynamically creating empty tables, Stripe currently has using__tablename variables to disable these. So for customers that have these set to false, nothing changes.
  - For customers that have these set to true, and the table exists, nothing changes. 
  - For customers that have these set to true, and the table doesn’t exist, we are resolving an error.
  - For tables that aren’t covered by the variables, if they don’t exist, they caused errors in previous versions. Now they won’t
- Xero
  - Instead of dynamically creating empty tables, xero currently has using__tablename variables to disable these (2). So for customers that have these set to false, nothing changes.
  - For customers that have these set to true, and the table exists, nothing changes. 
  - For customers that have these set to true, and the table doesn’t exist, we are resolving an error.
  - For tables that aren’t covered by the variables, if they don’t exist, they caused errors in previous versions. Now they won’t
- Klaviyo 
  - If a table doesn’t exist, in previous versions the package would have failed. Now it creates an empty ones and works with those

Packages upstream of the above:
- Shopify holistic reporting
- social media reporting

- [x] Yes

### DIFF
--- a/macros/union_data.sql
+++ b/macros/union_data.sql
@@ -55,6 +55,7 @@
     {%- else -%}
     select 
         cast(null as {{ dbt.type_string() }}) as _dbt_source_relation
+    limit 0
     {%- endif -%}
 
 {%- elif var(union_database_variable, none) -%}
@@ -81,6 +82,7 @@
     {%- else -%}
     select 
         cast(null as {{ dbt.type_string() }}) as _dbt_source_relation
+    limit 0
     {%- endif -%}
 
 {%- else -%}
@@ -97,6 +99,7 @@
 {%- else -%}
     select 
         cast(null as {{ dbt.type_string() }}) as _dbt_source_relation
+    limit 0
 {%- endif -%}
 {%- endif -%}
 
--- a/macros/union_data.sql
+++ b/macros/union_data.sql
@@ -53,6 +53,7 @@
     {%- if relations != [] -%}
         {{ dbt_utils.union_relations(relations) }}
     {%- else -%}
+    {% if execute and not var('remove_empty_table_warnings', false) -%}{{ exceptions.warn("\n\nWarning: The " ~ table_identifier ~ " table was not found in your destination. The package will create a completely empty staging model as to not break downstream transformations.\n") }}{% endif -%}
     select 
         cast(null as {{ dbt.type_string() }}) as _dbt_source_relation
     limit 0
@@ -80,6 +81,7 @@
     {%- if relations != [] -%}
         {{ dbt_utils.union_relations(relations) }}
     {%- else -%}
+    {% if execute and not var('remove_empty_table_warnings', false) -%}{{ exceptions.warn("\n\nWarning: The " ~ table_identifier ~ " table was not found in your destination. The package will create a completely empty staging model as to not break downstream transformations.\n") }}{% endif -%}
     select 
         cast(null as {{ dbt.type_string() }}) as _dbt_source_relation
     limit 0
@@ -97,6 +99,7 @@
     select * 
     from {{ var(default_variable) }}
 {%- else -%}
+    {% if execute and not var('remove_empty_table_warnings', false) -%}{{ exceptions.warn("\n\nWarning: The " ~ table_identifier ~ " table was not found in your destination. The package will create a completely empty staging model as to not break downstream transformations.\n") }}{% endif -%}
     select 
         cast(null as {{ dbt.type_string() }}) as _dbt_source_relation
     limit 0
--- a/macros/union_data.sql
+++ b/macros/union_data.sql
@@ -1,4 +1,4 @@
-{% macro union_data(table_identifier, database_variable, schema_variable, default_database, default_schema, default_variable, union_schema_variable='union_schemas', union_database_variable='union_databases') -%}
+{%- macro union_data(table_identifier, database_variable, schema_variable, default_database, default_schema, default_variable, column_macro=[{}], union_schema_variable='union_schemas', union_database_variable='union_databases') -%}
 
 {{ adapter.dispatch('union_data', 'fivetran_utils') (
     table_identifier, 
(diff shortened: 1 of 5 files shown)


fivetran-jamie:
in a way the column_macro stuff isn't the 100% necessary within the context of our source packages, as the non-tmp staging model will call `fill_staging_column` itself and do the null-casting....

macros/union_data.sql
```diff
@@ -7,82 +7,102 @@
...
-    {% endfor %}
-
-    {{ dbt_utils.union_relations(relations) }}
-
-{% elif var(union_database_variable, none) %}
+    {%- if relations != [] -%}
+        {{ dbt_utils.union_relations(relations) }}
+    {%- else -%}
+    select 
+        {% if column_macro != [{}] %}{{ fivetran_utils.fill_staging_columns(source_columns=[], staging_columns=column_macro) }},{% endif -%}
```
fivetran-joemarkiewicz: As I stated in the [shopify_source review](https://github.com/fivetran/dbt_shopify_source/pull/59), I think this is possibly duplicative. I think we can remove the column_macro since we can gain comfort that the staging model will ensure success and parity of fields.

Additionally, this results in less maintenance across the packages!

macros/union_data.sql
```diff
@@ -7,82 +7,102 @@
...
-    
-    {% endif %}
+    {%- endfor -%}
 
-    {% endfor %}
+    {%- if relations != [] -%}
+        {{ dbt_utils.union_relations(relations) }}
+    {%- else -%}
+    select 
+        {% if column_macro != [{}] %}{{ fivetran_utils.fill_staging_columns(source_columns=[], staging_columns=column_macro) }},{% endif -%}
```
fivetran-joemarkiewicz: Same comment as above.

fivetran-joemarkiewicz (commented):
@fivetran-jamie thanks for working through this unique issue and finding a solution that can be easily scaled out to all our packages! I did have one comment following my review of the shopify_source package and that is to consider removing the `column_macro` since this is not entirely necessary.

Let me know once that has been updated and I can do a final review.

macros/union_data.sql
fivetran-jamie: removed

macros/union_data.sql
fivetran-jamie: removed

fivetran-sheringuyen:
Overall, this is a really neat solution @fivetran-jamie! 

As a reviewer, I personally felt very easily overwhelmed/lost with this macro over all so thank you for taking the time to review it with me. (But it makes me consider if we should create a separate issue to provide like working examples of these more complex macros so that it doesn't require the review/explanation again 🤔 ). 

I'm wondering if in addition to making sure the test doesn't fail, that we should output some kind of messaging to ensure that the user knows that their listed tables were not actually found and have yielded an empty table(s). I'm just thinking that previously, they would have known due to a failure in the macro and could have had an option to fix it on their side. 

And a side note, once this branch merges with `feature/unit-testing-phase1` buildkite should pass!

fivetran-sheringuyen (approved):

fivetran-jamie:
i like that idea @fivetran-sheringuyen ! do you know if it's possible to add a custom message to a test output? 


also should i change the base branch of this PR to `feature/unit-testing-phase1`? or merge `feature/unit-testing-phase1` into my working branch? or wait for `feature/unit-testing-phase1` to merge into the release branch?  or something else altogether....

macros/union_data.sql
```diff
@@ -22,67 +22,88 @@
...
-
-    {% endfor %}
-
-    {{ dbt_utils.union_relations(relations) }}
-
-{% elif var(union_database_variable, none) %}
+    {%- if relations != [] -%}
+        {{ dbt_utils.union_relations(relations) }}
+    {%- else -%}
+    {% if execute and not var('remove_empty_table_warnings', false) -%}{{ exceptions.warn("\n\nWarning: The " ~ table_identifier ~ " table was not found in your destination. The package will create a completely empty staging model as to not break downstream transformations.\n") }}{% endif -%}
```
fivetran-joemarkiewicz: Now that I think about this, is "Warning" the right word here? A warning to me indicates something needs to be looked into, this may not 🤷. Is there a less severe word we could possibly use here?

fivetran-joemarkiewicz: Additionally, what are your thoughts in providing the variable to disable in the message itself? Just so customers are directly aware of it and can turn it off without needing to inspect our docs?

fivetran-joemarkiewicz (commented):
@fivetran-jamie before approving this I still want to test it on the other packages with the union data functionality, but I did have a few comments below which I would like your input.

Additionally, will we need to add the README and DECISIONLOG entry you have in shopify_source to the other packages as well? My understanding is that we will, but wanted to confirm on your end before approving. This way we can make a plan for what needs to be updated.

macros/union_data.sql
fivetran-jamie: >Now that I think about this, is "Warning" the right word here? A warning to me indicates something needs to be looked into, this may not 🤷. Is there a less severe word we could possibly use here?

Yeah honestly i thought about "FYI" or "Heads up" lol

>Additionally, what are your thoughts in providing the variable to disable in the message itself? Just so customers are directly aware of it and can turn it off without needing to inspect our docs?

 I really like that idea!!

fivetran-jamie:
ah yep we would have to update the docs for all the packages in the PR description (i realized i missed klaviyo and added the upstream affected packages). we should also update the shopify transform package

macros/union_data.sql
```diff
@@ -22,67 +22,94 @@
...
-
-    {% set relations = [] %}
-
-    {% for database in var(union_database_variable) %}
-
-    {% set relation=adapter.get_relation(
+    {%- if relations != [] -%}
+        {{ dbt_utils.union_relations(relations) }}
+    {%- else -%}
+    {% if execute and not var('remove_empty_table_warnings', false) -%}
```
fivetran-joemarkiewicz: Could this maybe be misleading if customers believe this to be a native dbt variable? Should we consider renaming this to be `fivetran__remove_empty_table_warnings` or something like that?

macros/union_data.sql
```diff
@@ -22,67 +22,94 @@
...
+    {{ exceptions.warn("\n\nPlease be aware: The " ~ table_identifier ~ " table was not found in your destination. The package will create a completely empty staging model as to not break downstream transformations. To turn off these warnings, set the `remove_empty_table_warnings` variable to False.\n") }}
```
fivetran-joemarkiewicz: What are your thoughts on adding the package information to the warning? I tested this locally where I was doing this union behavior for three packages and found it to be quite confusing which package was doing what with the staging models.

I found we can add the following to add a bit more details.
```sql
{{ exceptions.warn("\n\nPlease be aware: The " ~ table_identifier ~ " table was not found in your destination. The " ~ default_schema ~ " package will create a completely empty staging model as to not break downstream transformations. To turn off these warnings, set the `remove_empty_table_warnings` variable to False.\n") }}
```

Results:
```zsh
Please be aware: The discount_code table was not found in your destination. The shopify package will create a completely empty staging model as to not break downstream transformations. To turn off these warnings, set the `remove_empty_table_warnings` variable to False.

**AND**

Please be aware: The user_history table was not found in your destination. The instagram_business_pages package will create a completely empty staging model as to not break downstream transformations. To turn off these warnings, set the `remove_empty_table_warnings` variable to False.
```

It isn't fool proof as the naming won't match exactly what we have named the package (ie. we don't call the instagram_business package instagram_business_pages), but it may be enough to add more context.

I am trying to future proof this and I foresee us using this across a lot more packages, and feel this could get overwhelming for customers if they are unaware where these tables are being used. Thoughts?

fivetran-joemarkiewicz (commented):
@fivetran-jamie this update is incredible and I think we should start to leverage this more widespread across our packages!

I have a few minor comments that I feel could possibly enhance the user experience based on my integration of a number of union macros in packages. Let me know your thoughts!

macros/union_data.sql
fivetran-jamie: yeah i was thinking of a fivetran prefix as well! i'll update

fivetran-jamie: ah now i remember -- i opted for no prefix at first to be consistent with the `has_defined_sources` variable we also reference in this macro (IMO, _that_ name is confusing and should have a `fivetran__` prefix as well, but i don't wanna break anything here 😬)

i still think adding the prefix to the new variable is the best move forward tho

macros/union_data.sql
fivetran-jamie: lmk whatcha think about what i added!

fivetran-joemarkiewicz: Looks great, I like the added details to the warning message!

macros/union_data.sql
fivetran-joemarkiewicz: Yeah let's flag this as something to update in the next breaking version of fivetran_utils. I like the update!

fivetran-joemarkiewicz (approved):
@fivetran-jamie great work on this PR! I just did my final review and have tested the following packages to ensure they work seamlessly on their current build of fivetran_utils and on the new one that will results from this update:
- Shopify
- Stripe
- Xero
- Instagram Business
- Klaviyo

I can confirm that they work without error and the updates highlighted here are functioning as intended. This looks great to move forward into the release branch!

PR #101, opened 2023-03-07
Feature/unit testing phase1

fivetran-sheringuyen:
 
Adds a dbt compile test to our Fivetran Utils package

- [ ] Yes
- [x] No (provide further explanation)
Integration testing for internal uses


fivetran-sheringuyen:
I will..squash before merge 😅

integration_tests/packages.yml
```diff
@@ -1,2 +1,11 @@
...
+  - package: fivetran/twitter_ads
+    version: [">=0.6.0", "<0.7.0"]
+  - package: fivetran/zendesk
+    version: [">=0.10.0", "<0.11.0"]
+  - package: fivetran/hubspot
+    version: [">=0.8.0", "<0.9.0"]
+  - package: fivetran/jira
+    version: [">=0.11.0", "<0.12.0"]
+  - package: fivetran/klaviyo
+    version: [">=0.5.0", "<0.6.0"]
```
fivetran-joemarkiewicz, line 11: Should we maybe make these ranges larger? I would want to make sure we are testing on the latest versions of them.

fivetran-joemarkiewicz (approved):
@fivetran-sheringuyen this is great and will really help us unsure that changes to this package have more coverage than just releasing and hoping a potential bug isn't introduced!

I have one small comment, but other than that - this looks great to go! I do want to hold off merging and releasing this until PR #100 is completed.

integration_tests/packages.yml
fivetran-sheringuyen: I think there's a couple of different things we could do here. 
1) Broaden the range as you are suggesting and having the upper bound be < 1.0.0
or
2) Having `revision: main` for the git reference. This ensures that it's always the latest version of main, although not necessarily a release. So there is a minor caveat here but for this context, I think it would be ok.

PR #106, opened 2023-03-23
exploring integration test schema cleanup 

fivetran-jamie:
 
- adds drop_schemas macro to be run in integration testing
- adds quote macro that i feel like we would use all the time

on shopify source

- [x] Yes
- [ ] No (provide further explanation) --> still need to


README.md
```diff
@@ -344,6 +356,19 @@ vars:
 * `variable` (required): The variable containing the calculated field `name` and `transform_sql`.
 
 ----
+
+### drop_schemas ([source](macros/drop_schemas.sql))
```
fivetran-sheringuyen: I would maybe give an examples of how it can be used. 

I can see two use cases --
1) at the end of buildkite runs like you mentioned
2) in the deployments.yml

macros/drop_schemas.sql
```diff
@@ -0,0 +1,31 @@
+{% macro drop_schemas(drop_target_schema=true) %}
```
fivetran-sheringuyen: i wonder if you can make this more flexible by adding a `fuzzy_schema_match` condition? 

Specifically for [this use case](https://github.com/fivetran/dbt_package_integrations/blob/main/macros/delete_integration_tests.sql). If a `fuzz_schema_match = 'LIKE '%integration_test%'` could be a param, this can allow us to leverage the fivetran utils package for this instead of having a custom macro in the integrations package.

fivetran-sheringuyen: I guess it doesn't have to be fuzzy lol it could just be something like `schema_where_like` . but probably depends on the use cases.

fivetran-sheringuyen (commented):
Awesome PR! Love the cross team collab here, hoping to see more of this in the future! 

Additionally, something that may be interesting to consider is setting expiration dates for tables like we currently do for analytics [here](https://github.com/fivetran/analytics/blob/main/dbt_ft_prod/macros/operations/z_dbt_schema_expiration.sql).

macros/drop_schemas.sql
```diff
@@ -0,0 +1,31 @@
...
+
```
fivetran-joemarkiewicz: Should we add the dispatch config to this macro as well? Additionally what are your thoughts on naming this something unique to ensure there is no overlap with matching macro names in users projects?

README.md
fivetran-joemarkiewicz: I second this. It will be helpful to understand how this is being used in practice.

macros/drop_schemas.sql
```diff
@@ -0,0 +1,31 @@
...
+{% if execute %}
+{% set results_list = results.columns[0].values() %}
+{% else %}
+{% set results_list = [] %}
+{% endif %}
+
+{% for schema_to_drop in results_list %}
+
+{{ run_query("drop schema if exists " ~ schema_to_drop ~ " cascade;") }}
```
fivetran-joemarkiewicz: Should we consider using the [drop_schema](https://docs.getdbt.com/reference/dbt-jinja-functions/adapter#drop_schema) adapter function?

fivetran-joemarkiewicz (changes requested):
@fivetran-jamie I have a few comments and suggestions. Additionally, would you be able to increase the package index and add a log to the CHANGELOG. Let me know if you have any questions.

macros/drop_schemas.sql
fivetran-jamie: oooh okay so twoish ideas: 
- maybe a `schema_patterns` argument that by default is `[target.schema]` so you can drop multiple schema patterns. so in this case you would provide `['integration_tests', 'integrations_tests']`. i realize this isn't what you were focusing on above but realized that's also a thing 😅 or you'd have to run the macro twice with different targets
- i opted for only adding the `%` wildcard as a suffix since that's in line with how dbt does custom schemas (ie appends `_schema_suffx` to the target). but yeah maybe that's too limiting.. i wonder if we add the `schema_patterns` variable, should the user have to add the `%`'s to the patterns? so like in your example, you could get it extra fuzzy and provide `schema_patterns=['%integration_test%', '%integrations_test%']` 
  - however, the default behavior could still align with dbt and be `schema_pattern=[target.schema ~ '%']`

we could also have the user provide a whole `where` clause like you mentioned, which would offer the most flexibility but fewer guardrails

cc @fivetran-joemarkiewicz

macros/drop_schemas.sql
fivetran-jamie: there's no face smack emoji on github but know that i would reply with that...we should definitely use that lol

README.md
fivetran-jamie: added!

macros/drop_schemas.sql
fivetran-jamie: yeah i think we should. added both a dispatch and different name

macros/drop_schemas.sql
fivetran-jamie: updated!

README.md
```diff
@@ -344,6 +356,33 @@ vars:
 * `variable` (required): The variable containing the calculated field `name` and `transform_sql`.
 
 ----
+
+### drop_schemas_automation ([source](macros/drop_schemas_automation.sql))
+This macro was created to be clean up the schemas in our integration test environments. It drops schemas that are `like` the `target.schema`. By default it will drop the target schema as well but this can be configured.
```
fivetran-joemarkiewicz:
```suggestion
This macro was created to help clean up the schemas in our integration test environments. It drops schemas that are `like` the `target.schema`. By default it will drop the target schema as well but this can be configured.
```

macros/drop_schemas_automation.sql
```diff
@@ -0,0 +1,30 @@
+{% macro drop_schemas_automation(drop_target_schema=true) %}
+    {{ return(adapter.dispatch('drop_schemas_automation')(drop_target_schema)) }}
+{%- endmacro %}
```
fivetran-joemarkiewicz: Out of curiosity, is there a reason this dispatch looks different from the one used in the quote macro? Should we be including the fivetran_utils dispatch in here?

I have always been fuzzy with my understanding of the dispatch protocol. However, I think we should structure both to be in line with the latest structure of the dbt-utils macros. [Here](https://github.com/dbt-labs/dbt-utils/blob/9507d6bcd6c04b73e174a9aaee11865f9387e096/macros/sql/star.sql#L1-L3) is what they have for the `star` macro.

I believe adding `fivetran_utils` will ensure that the macro will only be used when prefixed with fivetran_utils and could not be at risk of overriding a users own macro. This is what I got from their [documentation](https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch#for-package-maintainers), interested to get your input on this as well.

macros/quote_something.sql
```diff
@@ -0,0 +1,22 @@
+{%- macro quote(thing) -%}
+
+{{ adapter.dispatch('quote', 'fivetran_utils') (thing) }}
```
fivetran-joemarkiewicz: Similar question regarding the dispatch. Is there a reason this does not have the `return` statement?

macros/quote_something.sql
```diff
@@ -0,0 +1,22 @@
+{%- macro quote(thing) -%}
```
fivetran-joemarkiewicz: Not a critical comment, but I was not fully aware of what this macro was doing until I read your explanation in the README. Would a more appropriate name and argument be something along the lines of 
```sql
{%-database_quoting(db_name) -%}
```
I'm not entirely strong on the opinion that this should be renamed, but curious what your thoughts are.

macros/drop_schemas_automation.sql
```diff
@@ -0,0 +1,30 @@
...
+{% set results = run_query(fetch_list_sql) %}
+
+{% if execute %}
+    {% set results_list = results.columns[0].values() %}
+{% else %}
+    {% set results_list = [] %}
+{% endif %}
+
+{% for schema_to_drop in results_list %}
+    {% do adapter.drop_schema(api.Relation.create(database=target.database, schema=schema_to_drop)) %}
```
fivetran-joemarkiewicz, line 27: What are your thoughts on adding a cli message to let us know that the schema was successfully dropped?

Looking at the Shopify PR you opened, it is hard to see that this is even happening. I can confirm when looking at the data that his was successful, but it may be a better debugging experience to clearly call out in the logs that the schema was dropped. Thoughts?

fivetran-joemarkiewicz (changes requested):
@fivetran-jamie thanks for making the requested updates! When doing my final review I had some additional questions and suggestions around the previous changes and the new ones you applied. That being said, the code looks to function without issue! I was able do some detailed testing on my end and feel confident that following this round of updates we should be good to approve and release!

Let me know if you want to sync to discuss in more detail.

macros/drop_schemas.sql
fivetran-joemarkiewicz: Yeah I think for the current use case for this macro the way @fivetran-jamie opted to build it out works great! I could see a future where we provide more flexibility. 

However, for the current implementation I would prefer we keep it as is so we don't run into any fuzzy matching issues within our own warehouses. Especially as this will be running constantly as we work on updates across our packages. If we see interest in people wanting to leverage this outside of the current use case, we should definitely consider expanding the flexibility of the macro.

macros/quote_something.sql
fivetran-jamie: well the thing _could_ be a database name, but it could also be a column or table (or anything else you'd need to quote in SQL?) ... perhaps `sql_object` would be a better argument name? 

as for the macro name, i felt like simpler felt nice (and we don't add `db_*` to our other cross-database macros) but maybe something more descriptive like `wrap_in_quotes` would be better?

fivetran-joemarkiewicz: Good points! I really like the `wrap_in_quotes` suggestion.

macros/drop_schemas.sql
fivetran-jamie: cool i'll make a FR to include fuzzy matching after we merge this in!

macros/drop_schemas_automation.sql
fivetran-jamie: added

macros/quote_something.sql
fivetran-jamie: added `return`

macros/quote_something.sql
fivetran-jamie: cool updated the name of the macro (but left `thing` as the argument?)

macros/drop_schemas_automation.sql
fivetran-jamie: added a cli message!

macros/quote_something.sql
fivetran-jamie: renamed `thing` to `object_to_quote`

fivetran-joemarkiewicz (approved):
@fivetran-jamie thanks for making these updates! All the changes look good to me!

PR #114, opened 2023-04-28
bugfix/union-data-coverage

fivetran-joemarkiewicz:
 
This PR adjusts the `union_data` macro to ensure a source connection may be established regardless of unioning schemas or not. This is a mirror PR to PR #110 with the addition of handling the error which resulted from `v0.4.5` where the `default_schema` name did not always map to the name of the source.yml in the package. This caused compilation errors for customers and **MUST** be avoided and addressed.

Unfortunately, the only way to address this is to hard code the exceptions and performing a proper mapping exercise. As this behavior is only present in two packages, it is not too much of a risk, but this can be expanded in the future to include more as we roll out this feature. 

A few other approaches I took include the following:
- Adding a new argument to the macro that allows the package to provide the actual source name (regardless if it matches the default schema name). However, this did not prove to be viable as this would only be available in the latest versions of the impacted packages. Therefore, customers who have pinned versions of the packages would experience failures which is not ideal.
- Using a different argument from the original macro to obtain the source name. This was inconclusive and default_schema name was consistently the best option
  - This did make me think that we should **always** make the source name the default schema name in our packages. At least for consistency.
- Performing a trip or some manipulation procedure on the default_schema name. This proved to be more "hacky" than the exception mapping approach I took.

Very much open to other ideas if any have them.

This is not a new macro but rather an expansion of the existing `union_data` macro.

This update mainly impacts packages using the union_data macro. As such, this version of fivetran_utils was tested on **all** of the following packages:
- dbt_shopify_source
- dbt_klaviyo_source
- dbt_stripe_source
- dbt_xero_source
- dbt_quickbooks_source
- dbt_linkedin_pages_source
- dbt_instagram_business_source
- dbt_twitter_organic_source
- dbt_facebook_pages_source

Additional packages this was tested on include:
- dbt_fivetran_log
- dbt_zendesk
- dbt_shopify_holistic_reporting

- [ ] Yes
The README updates were already applied in the previous PR linked above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# dbt_fivetran_utils v0.4.7
+## Feature Updates
+- Update to the `union_data` macro to ensure a source connection is established when just one schema is being used with the macro. 
+    - The previous build did allow the macro to read data from the correct location; however, it did not leverage the source. Therefore, the models using this macro would result in not having a source connection. This has since been updated in this release and all versions of this macro (whether unioning or not) will attempt to leverage the source function.
+- In addition to the above update, a specific code adjustment was provided for the LinkedIn Company Pages and Instagram Business Pages packages as the sources are not named the same as their default schema. This was the core issue that resulted in dbt_fivetran_utils v0.4.5 being rolled back. This has now been addressed and may be expanded in the future for any other packages with mismatching source/default_schema names.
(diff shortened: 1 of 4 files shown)


fivetran-joemarkiewicz:
Closing this as PR #118 covered this as well

PR #118, opened 2023-06-01
Feature/testing full coverge

fivetran-joemarkiewicz:
### Feature Updates
- Update to the `union_data` macro to ensure a source connection is established when just one schema is being used with the macro. 
    - The previous build did allow the macro to read data from the correct location; however, it did not leverage the source. Therefore, the models using this macro would result in not having a source connection. This has since been updated in this release and all versions of this macro (whether unioning or not) will attempt to leverage the source function.
- In addition to the above update, a specific code adjustment was provided for the LinkedIn Company Pages and Instagram Business Pages packages as the sources are not named the same as their default schema. This was the core issue that resulted in dbt_fivetran_utils v0.4.5 being rolled back. This has now been addressed and may be expanded in the future for any other packages with mismatching source/default_schema names.
    - For additional context, this is the approach taken as renaming the sources could result in incompatible versions of base package and fivetran_utils. Particularly, this could result in package users having a different experience based on the version of their base package. Which is not the behavior we would prefer for this new feature.

### Under the Hood
- The `union_data` macro has been adjusted to leverage the identifier variable and the true relation when running on a single schema. This ensures a more accurate use of the macro and the ability for the macro to work as expected within the Fivetran integration test pipeline.
- The `.buildkite` folder has been overhauled to ensure full integration test coverage over **all** Fivetran dbt packages that reference the dbt_fivetran_utils package. The new pipeline updates will ensure the fivetran_utils release version will succeed and not cause errors on any Fivetran dependent package prior to release.

## Question for the Reviewer
The Databricks tests seems to be a fail point for the integration tests due to our allocated cluster and worker size. What are your thoughts on possibly doing a trimmed down version of these tests? Alternatively, we can investigate the cost of expanding our cluster size to handle the workload of these integration tests. Curious what your opinion is.

## Tips for Reviewing
I was able to test this integration pipeline locally by creating a new file `test_run_models.sh` (added to the `.gitignore`) which is a copy of the `run_standard_models.sh` file and running that from the fivetran_utils directory. See my setup below for how I ran this file to test the functionality (please note you can only test this on one warehouse at a time).
>**Note**: This will edit your local `profiles.yml` file. It will just change the schema for each target. I usually can just command + z to go back, but something to take not of if you use this method to test.

image

## If this PR introduces a new macro, how did you test the new macro? 

The updates have been tested using the new integration tests pipeline. See buildkite for more details.

## If this PR introduces a modification to an existing macro, which packages is the macro currently present in and what steps were taken to test compatibility across packages?

In addition to adding the new integration tests pipeline, this macro also updates the `union_data` macro. The old version of the macro was actually causing failures in the integration tests pipeline. I have since been able to confirm that the updates to the macro ensure a valid connection to the source relation, **and** allow for a source connection to the source tables. See the screenshot below for details which is the result of running the new macro for the shopify package and identifying the source is established properly.

image

## Did you update the README to reflect the macro addition/modifications?

- [X] No (there were no new macros added and the README was not needed to be updated.)


.buildkite/scripts/run_standard_models.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+## remove failure for testing
+set -euo pipefail
+
```
fivetran-sheringuyen, line 4: Holy smokes this is a long pipeline file!

Have you thought about breaking it up into multiple files? I'm not exactly sure how it will look like on the UI, but I'm guessing it would look the same. May be nice to just have a pipeline file for each of the categories - Ad Reporting, App Reporting...etc. Of course, I don't think it's necessary but may help with navigation and maintenance.

We have a few different checks that we are running and have split them into different files in our internal analytics repo. The [different pipelines](https://github.com/fivetran/analytics/tree/main/.buildkite) and how they look like in [Buildkite](https://buildkite.com/fivetran/analytics-test-pipeline/builds/2205).

.buildkite/scripts/run_standard_models.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+## remove failure for testing
```
fivetran-sheringuyen: I checked the blame, can't for the life of me quite remember what this comment is for lol I _think_ it had something to do with adding in this `-euo` flag for any failures to fail the run right away rather than continuing to execute. But we could probably just remove the comment or make it clearer

.buildkite/scripts/run_standard_models.sh
```diff
@@ -0,0 +1,52 @@
...
+cp integration_tests/ci/sample.profiles.yml ~/.dbt/profiles.yml
+
+db=$1
+echo `pwd`
+cd integration_tests
+dbt deps ## Install all packages needed
+
+shift ## Skips the first argument (warehouse) and moves to only looking at the data model arguments
+
+for model in "$@" ## Iterates over all non warehouse arguments
```
fivetran-sheringuyen: If I'm understanding correctly, the "model" variable that you're using to iterate with is actually a "package"? I would suggest using "package" just so that's clear here.

.buildkite/scripts/run_standard_models.sh
```diff
@@ -0,0 +1,52 @@
...
+do
+    echo -e "\ncompiling "$model"\n"
+    cd dbt_packages/$model/integration_tests/
+    dbt deps
+    cp ../../../packages_ft_utils_override.yml packages.yml
+    dbt deps
+    if [ "$model" = "linkedin" ]; then
+        value_to_replace=$(grep ""$model"_ads_schema:" dbt_project.yml | awk '{ print $2 }')
```
fivetran-sheringuyen: Was it intentional to use "$model" here instead of linkedin?

.buildkite/scripts/run_standard_models.sh
```diff
@@ -0,0 +1,52 @@
...
+        perl -i -pe "s/(schema: |dataset: ).*/\1$value_to_replace/" ~/.dbt/profiles.yml
+    elif [ "$model" = "ad_reporting" ]; then
+        value_to_replace=$(grep "google_ads_schema:" dbt_project.yml | awk '{ print $2 }')
```
fivetran-sheringuyen, line 33: In the `pipeline.yml` it looks like there's a `Run Tests - Ad Reporting Roll Up Package` as well as individual runs in `Ad Packages 3` for google ads. Is there a reason we're only replacing google_ads? 

Similar question for the app reporting section down below.

.buildkite/scripts/run_standard_models.sh
```diff
@@ -0,0 +1,52 @@
...
+shift ## Skips the first argument (warehouse) and moves to only looking at the data model arguments
+
+for model in "$@" ## Iterates over all non warehouse arguments
+do
+    echo -e "\ncompiling "$model"\n"
+    cd dbt_packages/$model/integration_tests/
+    dbt deps
+    cp ../../../packages_ft_utils_override.yml packages.yml
+    dbt deps
+    if [ "$model" = "linkedin" ]; then
```
fivetran-sheringuyen: Would you mind taking a second to explain to me what should be happening in these if statements?

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+# dbt_fivetran_utils v0.4.7
+[PR #118](https://github.com/fivetran/dbt_fivetran_utils/pull/118) includes the following updates.
+## Feature Updates
+- Update to the `union_data` macro to ensure a source connection is established when just one schema is being used with the macro. 
+    - The previous build did allow the macro to read data from the correct location; however, it did not leverage the source. Therefore, the models using this macro would result in not having a source connection. This has since been updated in this release and all versions of this macro (whether unioning or not) will attempt to leverage the source function.
+- In addition to the above update, a specific code adjustment was provided for the LinkedIn Company Pages and Instagram Business Pages packages as the sources are not named the same as their default schema. This was the core issue that resulted in dbt_fivetran_utils v0.4.5 being rolled back. This has now been addressed and may be expanded in the future for any other packages with mismatching source/default_schema names.
```
fivetran-sheringuyen: Link to code/lines of code?

integration_tests/packages.yml
```diff
@@ -1,11 +1,80 @@
...
-    version: [">=0.10.0", "<1.0.0"]
+  - package: fivetran/ad_reporting
+    version: [">=1.0.0", "<2.0.0"]
+
+  - package: fivetran/shopify_holistic_reporting
+    version: [">=0.1.0", "<1.0.0"]
+
+  ## To be switched after the social media integration tests is updated
+  # - package: fivetran/social_media_reporting
+  #   version: [">=0.1.0", "<1.0.0"]
```
fivetran-sheringuyen, line 10: Obligatory reminder to do what you need to do here! Were you planning on cutting the release after this utils version is released?

I did not see an issue out for updating the integration testing in this package - may be good to have one so that it doesn't get lost.

integration_tests/packages.yml
```diff
@@ -1,11 +1,80 @@
...
+  - package: fivetran/jira
+    version: [">=0.1.0", "<1.0.0"]
+  - package: fivetran/github
+    version: [">=0.1.0", "<1.0.0"]
+  - package: fivetran/asana
+    version: [">=0.1.0", "<1.0.0"]
+  - package: fivetran/zendesk
+    version: [">=0.1.0", "<1.0.0"]
+  - package: fivetran/intercom
+    version: [">=0.1.0", "<1.0.0"]
```
fivetran-sheringuyen, line 80: I noticed that `salesforce_formula_utils` is not included here. Are there any plans for testing around those packages? Do you think it would be something you/your team may explore including in the future?

macros/union_data.sql
```diff
@@ -92,16 +92,34 @@
...
+    {% else %}
+        {# In order for this macro to effectively work within upstream integration tests (mainly used by the Fivetran dbt package maintainers), this identifier variable selection is required to use the macro with different identifier names. #}
+        {% set identifier_var = default_schema + "_" + table_identifier + "_identifier"  %}
+        {%- set relation.value=adapter.get_relation(
+            database=source(default_schema, table_identifier).database,
+            schema=source(default_schema, table_identifier).schema,
+            identifier=var(identifier_var, table_identifier)
+        ) -%}
+    {% endif %}
+{%- set table_exists=relation.value is not none -%}
```
fivetran-sheringuyen, line 122: Has this macro also been reviewed by someone on your team? Is this the same macro that Jamie originally created but was rolled back with this added exceptions logic? I think specifically for this section of the PR, I'd like a second set of eyes on this maybe @fivetran-jamie?

.buildkite/scripts/run_standard_models.sh
```diff
@@ -0,0 +1,52 @@
...
+. venv/bin/activate
+pip install --upgrade pip setuptools
+pip install -r integration_tests/requirements.txt
+mkdir -p ~/.dbt
+cp integration_tests/ci/sample.profiles.yml ~/.dbt/profiles.yml
+
+db=$1
+echo `pwd`
+cd integration_tests
+dbt deps ## Install all packages needed
```
fivetran-sheringuyen, line 18: I noticed that you're using the same `run_standard_models.sh` file for all of your pipeline concurrent steps. I think if in the future, we wish to optimize this process -- it may be a good idea to not `dbt deps` all packages and only packages that will be run in that step.

fivetran-sheringuyen (commented):
Hey @fivetran-joemarkiewicz ! Great job on this PR, I know it wasn't easy at all. 

I left some initial comments to guide our conversation for tomorrow. I still have to review the `union_data` macro and would like to understand particular pieces of the script so I can more thoroughly test the pipeline. 

As for speeding up Databricks, I know you've tried increasing the cluster (and perhaps other components like workers?) but I wonder if we could automate that process by leveraging [Databricks' autoscaling feature](https://www.databricks.com/blog/2018/05/02/introducing-databricks-optimized-auto-scaling.html). I'm not sure if that's a particular feature that must be enabled but it looks like there may be some functionality for autoscaling. The way I'm interpreting it is that when utilization is high, the cluster will scale up resources so that it can execute compute tasks quicker, in times of lower utilization, it automatically will resize the cluster down. From the article, it looks like there may be ways to track utilization capacities for upper and lower bounds. This may be something worth exploring to inform the bounds of the autoscaler.

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: I like this idea and meant to keep them in one file for the time being as I didn't want to add noise that may cause false failures. Now that we know the pipeline works, I think it is worth exploring to make maintenance easier down the road.

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: Sounds good and will remove in next commit!

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: I _think_ I used model because at one point I had two for loops and had package in the second one. Now that I condensed it into one loop I agree that using package would be much more consistent and less confusing.

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: "$model" is needed because that is how we fetch the component of the for loop. Here we are just checking if the component of the loop is equal to linkedin.

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: Yeah it is because the roll up packages have a list of schema variables all with the same value. Instead of iterating through all of them, we only need to check one and pull it for the profiles.yml. Therefore, I just picked one (Google in this instance) and used that as the origin for the profiles.yml.

image

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: Sure thing! The if statements are essentially needed because the packages that require a custom conditional are not structured in the same way as all the others. For example, we call the linkedin package `linkedin` however, all of the variables are named `linkedin_ads`. This shifts the structure of how the script code works because it by default will just use the package name to search for the schema variable. We need to make a small modification to ensure it works for linkedin then in this case. The same goes for the other packages with conditionals as well.

CHANGELOG.md
fivetran-joemarkiewicz: Good point, will link in next commit.

integration_tests/packages.yml
fivetran-joemarkiewicz: Yup, I want to make an update to social media before we even cut this. The changes are actually combined in [this PR](https://github.com/fivetran/dbt_social_media_reporting/pull/6/files#diff-20d7b76f07ce8c8ffc1405b865171ba1614babd41e7e357f72bcba11d2887fbe). Will have a review and release before we roll this out.

integration_tests/packages.yml
fivetran-joemarkiewicz: I actually did a quick assessment of that package for this coverage. It turns out the [only macro we use from fivetran_utils ](https://github.com/search?q=repo%3Afivetran%2Fdbt_salesforce_formula_utils%20fivetran_utils&type=code)in that package is the `fivetran_utils.drop_schemas_automation` macro and it is only used in our integration tests pipeline. Further, the package is structured quite differently from all the others that it would require a pretty customer testing pipeline. 

After doing the assessment I felt it was more work to get that package included than the value it would provide. I do think we should consider testing this further in the future, but felt it wasn't a necessary component for this PR.

macros/union_data.sql
fivetran-joemarkiewicz: @fivetran-jamie and I did chat about this informally during the offsite. That being said, I would agree with you that a second set of eyes on the union_data piece especially would be ideal.

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: I completely agree, I just opted not to do it this time around since that didn't seem to be a blocker and would require some string matching that I was not willing to dive into in this PR haha.

Additionally, the idea behind `run_standard_models.sh` is that all the standard runs (included in this PR) would be held in that file. My idea is to then expand this to other .sh files that test different components. But that is for a different PR 😄

.buildkite/scripts/run_standard_models.sh
fivetran-sheringuyen: Not resolving so that we can keep in PR for later reference.

.buildkite/scripts/run_standard_models.sh
fivetran-sheringuyen: Not resolving to keep for later reference.

fivetran-sheringuyen (commented):
Testing locally:
- `run_standard_models.sh`  is working great across all DBs
- Docs run and serve working as expected on the buggy instance of `union_data` (e.g. shopify package)

After discussing my questions this morning, I feel like this PR is in a good place for merge and release contingent on Jamie's feedback re: union_data and some minor PR fixes. I will give this a final review once you incorporate some of the changes we've discussed in the PR.

I want to add on as a last comment to track for the future. This is a great feat to see all of the tests passing for all of our Fivetran Packages -- huge win!! 🥳  Some time down the line, it would be great to optimize the runs so that we are not brute force testing every single package every single time. Instead, a more targeted CI run process would be great. E.g. only run tests on new packages, only run tests on packages that are using a particular macro etc. The internal Architecture team is currently exploring the idea of more targeted and intentional testing on our prod repos - hopefully we can work together to get something working in the Fivetran dbt packages using a similar framework.

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: I did a quick assessment and identified that this will be much more of a lift than I would like to incorporate within this release. As such, I created a [FR](https://github.com/fivetran/dbt_fivetran_utils/issues/119) so we can track this in the future and properly plan for this update.

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: [FR](https://github.com/fivetran/dbt_fivetran_utils/issues/120) opened to address this in the future.

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: Updated

.buildkite/scripts/run_standard_models.sh
fivetran-joemarkiewicz: Removed

CHANGELOG.md
fivetran-joemarkiewicz: Added relevant links

fivetran-joemarkiewicz:
@fivetran-sheringuyen just applied your updates (killed BuildKite to wait for @fivetran-jamie review in case there are other updates), but will be sure to re-run before/after approval and release.

macros/union_data.sql
```diff
@@ -92,16 +92,38 @@
     {%- endif -%}
 
 {%- else -%}
-    {%- set relation=adapter.get_relation(
-        database=var(database_variable, default_database),
-        schema=var(schema_variable, default_schema),
-        identifier=var(default_schema ~ '_' ~ table_identifier ~ '_' ~ 'identifier', table_identifier)) -%}
-
-{%- set table_exists=relation is not none -%}
+    {% set exception_schemas = {"linkedin_company_pages": "linkedin_pages", "instagram_business_pages": "instagram_business"} %}
```
fivetran-jamie, line 95: are these the only `exception_schemas` across ALL of our packages? or just those that use `union_data`? if it's the latter, we'll need to remember to check this for any packages we add unioning-ability too. 

wondering again about maintaining a spreadsheet (or seed file in fivetran_utils?) with info like this on each package

macros/union_data.sql
```diff
@@ -92,16 +92,38 @@
...
+    {% set relation = namespace(value="") %}
+    {% if default_schema in exception_schemas.keys() %}
+        {% for corrected_schema_name in exception_schemas.items() %}   
+            {% if default_schema in corrected_schema_name %}
+                {# In order for this macro to effectively work within upstream integration tests (mainly used by the Fivetran dbt package maintainers), this identifier variable selection is required to use the macro with different identifier names. #}
+                {% set identifier_var = corrected_schema_name[1] + "_" + table_identifier + "_identifier"  %}
+                {%- set relation.value=adapter.get_relation(
+                    database=source(corrected_schema_name[1], table_identifier).database,
+                    schema=source(corrected_schema_name[1], table_identifier).schema,
+                    identifier=var(identifier_var, table_identifier)
```
fivetran-jamie, line 105: this works for now, but this reallllly wouldn't work for something like the [apple store](https://github.com/fivetran/dbt_apple_store_source/blob/main/models/src_apple_store.yml), how do we feel about that?

macros/union_data.sql
```diff
@@ -92,16 +92,38 @@
...
+            {% endif %}
+        {% endfor %}
+    {% else %}
+        {# In order for this macro to effectively work within upstream integration tests (mainly used by the Fivetran dbt package maintainers), this identifier variable selection is required to use the macro with different identifier names. #}
+        {% set identifier_var = default_schema + "_" + table_identifier + "_identifier"  %}
+        {# Unfortunately the Twitter Organic identifiers were misspelled. As such, we will need to account for this in the model. This will be adjusted in the Twitter Organic package, but to ensure backwards compatibility, this needs to be included. #}
+        {% if var(identifier_var, none) is none %} 
+            {% set identifier_var = default_schema + "_" + table_identifier + "_identifer"  %}
+        {% endif %}
+        {%- set relation.value=adapter.get_relation(
```
fivetran-jamie, line 116: also why relation.value instead of relation?

fivetran-jamie (commented):
union_data looks pretty good but i do have concerns for sustainability (which maybe is for another time, another PR). but very impressed by this PR though! 

this also really highlights for me the need to consolidate information around each package, as fivetran_utils affects them all and makes assumptions left and right (or requires a bunch of arguments/hard-coding to figure out how to treat them)

fivetran-sheringuyen (approved):
LGTM! Keeping unresolved comments for visibility in the future and have resolved PR fix requests.

macros/union_data.sql
fivetran-joemarkiewicz: I only checked for the packages that currently leverage the `union_data` macro. I agree that this will be something we will need to take into consideration prior to using the macro in a package.

To also note, if a package is an exception (like the two listed here) it will fail. So we will be able to catch it pretty easily when implementing the macro.

macros/union_data.sql
fivetran-joemarkiewicz: Great question, I had to use namespaces in order to ensure the relation value would be accessible outside of a for loop. Since we set the namespace variable [here](https://github.com/fivetran/dbt_fivetran_utils/blob/6fd189a87eeb6370b1697e562fdac6b15e288270/macros/union_data.sql#L96), we then need to access the value by specifying it in the variable call.

macros/union_data.sql
fivetran-joemarkiewicz: This "technically would" work for a package like apple store. We would just need to add apple store to the exception dictionary. I agree that this isn't a great fool proof solution, but for right now it provides the coverage needed to ensure the updates work as intended.

fivetran-jamie: the corrected_schema for apple_store would be `itunes_connect`, but we still call the _source_ the `apple_store` (so this would compile to `source(itunes_connect, table_name)` and fail right?  we also don't add prefixes to the table-identifier variables in that package 🤔 ....regardless this solution works for now!

macros/union_data.sql
fivetran-jamie: ahh right the namespace! good to know

fivetran-jamie (approved):

integration_tests/packages.yml
fivetran-joemarkiewicz: Holding off on making this update until we release. The Social Media update is actually dependent on these changes to have buildkite pass. Therefore, we will release this update and then release a new version of social media. We can then include the official social version in the next update to fivetran utils

PR #129, opened 2023-11-14
add and incorporate macro for turning jsons (and only jsons) into strings

fivetran-jamie:
 

- Adds new macro, `get_json_columns_in_relation` to return a list of columns that are of type JSON (on BigQuery only, as we haven't rolled out JSON support in other Fivetran destinations). 
- Includes call to `get_json_columns_in_relation` in `fill_staging_columns`, so that each JSON field gets converted to a json-formatted STRING. This is necessary because only some connectors have JSONs currently and downstream json functions do not work seamlessly with the actual JSON-type columns.
  - `fill_staging_columns` was chosen as this is called in every (most?) source package model. therefore, we don't have to mass update a bunch of packages. however, we want this to be right and not break anything! 

## get_json_columns_in_relation
I first tested `get_json_columns_in_relation` by directly calling it in a local test model.

the model i ran it _against_ was just a select * from Joe's seed file 
`tiktok_ads_source_integration_tests_2.tiktok_adgroup_history_data`, which contains a JSON field called `AGE_GROUPS`. 

image

When running, the test model, i was able to get the macro to return and output `['AGE_GROUPS']`

## fill_staging_columns
I tested the incorporation of `get_json_columns_in_relation` into `fill_staging_columns` by running the tiktok ads source package (specifically from the `integration_tests` folder) locally. 

With `tiktok_adgroup_history_data` as the source, the model where trouble would pop up was `stg_tiktok_ads__ad_group_history`, specifically on the line where we try to coalesce `age_groups` and an actual-string field, `age`.  
image

With my additions, the fields CTE in `stg_tiktok_ads__ad_group_history` compiles to the following 
```sql
fields as (

    select
       
    
    action_days
   
        as 
    
    action_days
    
, 
    
    adgroup_id
    
        as 
    
    adgroup_id

# ................ cutting stuff out that's not relevant .......................... #

, 
    cast(null as string) as 
    
    age
    
 , 
    TO_JSON_STRING( 
    
    age_groups
    
 )
        as 
    
    age_groups
    
, 
    
    languages
   
        as 
    
    languages
    

, cast('' as string) as source_relation

    from base
), 
```

And so the coalesce of `age_groups` and `age` in the final CTE does not cause an error and the model runs successfully. 

and the data looks OK in bigquery
image

Honestly we should test this with other packages as well -- ones with JSONs and not -- given the far-reaching nature of this macro change. perhaps we could have some customers try this out

- [x] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ dispatch:
   - [Cross-database compatibility](#cross-database-compatibility)
     - [array\_agg (source)](#array_agg-source)
     - [ceiling (source)](#ceiling-source)
+    - [get\_json\_columns\_in\_relation (source)](#get_json_columns_in_relation-source)
     - [first\_value (source)](#first_value-source)
     - [json\_extract (source)](#json_extract-source)
     - [json\_parse (source)](#json_parse-source)
@@ -81,6 +82,7 @@ dispatch:
     - [remove\_prefix\_from\_columns (source)](#remove_prefix_from_columns-source)
     - [source\_relation (source)](#source_relation-source)
     - [union\_data (source)](#union_data-source)
+      - [Union Data Defined Sources Configuration](#union-data-defined-sources-configuration)
     - [union\_relations (source)](#union_relations-source)
   - [Variable Checks](#variable-checks)
     - [empty\_variable\_warning (source)](#empty_variable_warning-source)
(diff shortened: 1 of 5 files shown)


fivetran-jamie:
keeping this in our back pocket for now. 

if we want to move forward with this PR we should add a variable that is a boolean, telling us whether or not to convert back to strings or maintain the json type

PR #130, opened 2023-11-27
Bug/databricks get url parameter

fivetran-catfritz:
 
- Added macro `extract_url_parameter` to create special logic for Databricks instances not supported by `dbt_utils.get_url_parameter()`. The macro uses `dbt_utils.get_url_parameter()` for default, non-Databricks targets. See README for more details.

- Tested in linkedin_source on a Databricks SQL instance, which did not give errors like before. 
  - Stems from https://github.com/fivetran/dbt_linkedin_source/issues/55. 
- Also passes buildkite.

- Does not modify existing `fivetran_utils` macro, but is meant to replace `dbt_utils.get_url_parameter()`. 
- The affected packages tend use this macro at different stages, so the list is a result of a search for `dbt_utils.get_url_parameter`:
  - dbt_facebook_ads
  - dbt_google_ads_source
  - dbt_google_ads
  - dbt_linkedin_source
  - dbt_microsoft_ads
  - dbt_pinterest_source
  - dbt_reddit_ads
  - dbt_snapchat_ads
  - dbt_tiktok_ads_source
  - dbt_twitter_source
- Copy-cat issues have been created in these packages.

- [x] Yes


integration_tests/packages.yml
```diff
@@ -2,12 +2,8 @@ packages:
   - package: fivetran/ad_reporting
```
fivetran-joemarkiewicz, line 2: It looks like there are some upgrades we should apply to ensure we are testing the latest versions of all the packages.

image

macros/extract_url_parameter.sql
```diff
@@ -0,0 +1,20 @@
...
+
+
+{% macro default__extract_url_parameter(field, url_parameter) -%}
+
+{{ dbt_utils.get_url_parameter(field, url_parameter) }}
+
+{%- endmacro %}
+
+
+{% macro databricks__extract_url_parameter(field, url_parameter) -%}
```
fivetran-joemarkiewicz: It looks like in all our other macros that include a dispatch for databricks, we provide the `spark` dispatch:
- [json_parse](https://github.com/fivetran/dbt_fivetran_utils/blob/c536047a06153abfed46c272483bd768b2d34263/macros/json_parse.sql#L38)
- [try_cast](https://github.com/fivetran/dbt_fivetran_utils/blob/c536047a06153abfed46c272483bd768b2d34263/macros/try_cast.sql#L50)
- [timestamp_add](https://github.com/fivetran/dbt_fivetran_utils/blob/f7567295c8f406946dfdbcdd761affad7d7bb535/macros/timestamp_add.sql#L41)

Should we be using `spark` dispatch as opposed to `databricks`?

It looks like the `databricks` dispatch inherits the `spark` dispatch (see ref [here](https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch#adapter-inheritance)). With this, I believe we should leverage the `spark` dispatch so this will work regardless of a customer leveraging the spark or databricks adapter. Thoughts?

macros/extract_url_parameter.sql
```diff
@@ -0,0 +1,20 @@
...
+
+{%- set formatted_url_parameter = "'" + url_parameter + "=([^&]+)'" -%}
+nullif(regexp_extract({{ field }}, {{ formatted_url_parameter }}, 1), '')
```
fivetran-joemarkiewicz, line 18: So it looks like the only change is that you removed both `split()` macros that were initially found in the compiled version of the databricks macro. What was the split being used for before? Are we possibly not taking something into consideration with why the split was there in the first place?

fivetran-joemarkiewicz (changes requested):
@fivetran-catfritz great work on this PR! I just have a few small comments and suggestions I would like you to look at and address before approving. Let me know if you have any questions!

macros/extract_url_parameter.sql
fivetran-catfritz: Updated!

integration_tests/packages.yml
fivetran-catfritz: Updated!

macros/extract_url_parameter.sql
fivetran-catfritz: Summarizing our discussion from earlier:
- The regexp_extract() in the prior version was used to check for special characters, while the split() did the work of extracting the url parameter
- In this version, the regexp_extract() replaces the split() to do the work of extracting the url parameter, and it also automatically handles special characters.  
- I tested in our and the customer's environments, and it works for both our cases

fivetran-catfritz:
Thanks @fivetran-joemarkiewicz! I made the changes and added comments, so it's ready for re-review.

macros/extract_url_parameter.sql
fivetran-joemarkiewicz: Thanks for the explanation. This sounds good to me!

fivetran-joemarkiewicz (approved):
Thanks @fivetran-catfritz these changes look good to me! Let's hold off merging this to the live branch until we also incorporate the sql server updates as well.

If you have the time, would you be able to create a release branch for `release/v0.4.9`. We can then merge this branch into there, then the sql server updates into that branch as well before then rolling out the update to the live branch.

PR #133, opened 2023-12-05
Test/package lock removal

fivetran-joemarkiewicz:
Quick update to edit the integration tests to change the package-lock.yml instead of the packages.yml as this is needed for post dbt 1.7.0 testing.

### DIFF
--- a/.buildkite/scripts/run_standard_models.sh
+++ b/.buildkite/scripts/run_standard_models.sh
@@ -24,7 +24,8 @@ do
     echo -e "\ncompiling "$package"\n"
     cd dbt_packages/$package/integration_tests/
     dbt deps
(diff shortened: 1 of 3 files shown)


.buildkite/scripts/run_standard_models.sh
```diff
@@ -24,7 +24,8 @@ do
     echo -e "\ncompiling "$package"\n"
     cd dbt_packages/$package/integration_tests/
     dbt deps
-    cp ../../../packages_ft_utils_override.yml packages.yml
```
fivetran-jamie, line 27: so this `cp` is now excecuted via the print statements you added?

fivetran-jamie: does this mean we should remove the `packages_ft_utils_override.yml`?

PR #139, opened 2023-12-26
add fivetran_union_relations macro

fivetran-jamie:
 

Adds `fivetran_union_relations` macro. This was necessary for rolling union_data out to Zendesk, which has some source tables that have reserved-keywords for names (ie `timezone`). In the past, we've solved this error by aliasing the problematic source table, but there's no way to get `dbt_utils.union_relations` to do so (besides opening a PR on dbt utils or making our own version of the macro). 

We already _do_ have our own version of `union_relations` in this package, but as we learned [recently](https://github.com/fivetran/dbt_fivetran_utils/releases), using the same exact name as a macro in debt_utils can cause failures for customers with certain dispatch configs

This PR also adjusts union_data (in order to work with Netsuite, which is a weird package due to our support of both of its endpoints).
- new optional (none by default) `connector_table_name_override` field. this is to be used when the defined `source` table name is different from the _actual_ table name as it appears in the connector/warehouse. This is only a thing for Netsuite2 
  - Added a reference to a new global variable `use_table_name_identifer_override`. this was necessary to add for our integration tests (and customers) to still be able to use identifier  variables for tables in which `connector_table_name_override` is implemented. 

Ran it with zendesk, hubspot, and netsuite in their `feature/add-union-data` branches. moreover, i tested this fivetran_utils branch on Shopify_source, which has the union ability already. 

- just added a comment to the `union_relations` macro to deprecate it in the future 

Ran it with zendesk, hubspot, and netsuite in their `feature/add-union-data` branches. moreover, i tested this fivetran_utils branch on Shopify_source, which has the union ability already. 

- [x] Yes


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+# dbt_fivetran_utils v0.4.11
+
+## Feature Updates
+[PR #139](https://github.com/fivetran/dbt_fivetran_utils/pull/139) includes the following update:
+- Added the new `fivetran_union_relations()` macro.
+  - This was necessary for adding `union_data()` support to Zendesk the package, which works with some tables that have reserved-keywords for names (ie timezone).
+  - We heavily leveraged `dbt_utils.union_relations()` but adjusted it to work with these problematically named source tables.
+- Adjusted the `union_data()` macro to reference `fivetran_union_relations()` instead of `dbt_utils.union_relations()`.
+
+> Note: We already do have a macro calles `union_relations()` in this package, but as we learned [recently](https://github.com/fivetran/dbt_fivetran_utils/releases/tag/v0.4.10), using the same exact name as a macro in `dbt_utils` is not a good idea. Thus, `fivetran_utils.union_relations()` will be deprecated in favor of `fivetran_utils.fivetran_union_relations()` in the future. Currently it is used in only one package (Marketo).
```
fivetran-joemarkiewicz:
```suggestion
> Note: We already do have a macro called `union_relations()` in this package, but as we learned [recently](https://github.com/fivetran/dbt_fivetran_utils/releases/tag/v0.4.10), using the same exact name as a macro in `dbt_utils` is not a good idea. Thus, `fivetran_utils.union_relations()` will be deprecated in favor of `fivetran_utils.fivetran_union_relations()` in the future. Currently it is used in only one package (Marketo).
```

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+# dbt_fivetran_utils v0.4.11
+
+## Feature Updates
+[PR #139](https://github.com/fivetran/dbt_fivetran_utils/pull/139) includes the following update:
+- Added the new `fivetran_union_relations()` macro.
+  - This was necessary for adding `union_data()` support to Zendesk the package, which works with some tables that have reserved-keywords for names (ie timezone).
+  - We heavily leveraged `dbt_utils.union_relations()` but adjusted it to work with these problematically named source tables.
+- Adjusted the `union_data()` macro:
+  - Now references `fivetran_union_relations()` instead of `dbt_utils.union_relations()`.
+  - Now includes a new optonal argument, `connector_table_name_override`, which is the _actual_ name/identifier of the table as it appears in the connector schema. Currently, this is only used in our Netsuite2 data models, whose actual table names and defined `source` table names differ slightly due to our support of both the Netsuite1 and Netsuite2 endpoints.
```
fivetran-joemarkiewicz:
```suggestion
  - Now includes a new optional argument, `connector_table_name_override`, which is the _actual_ name/identifier of the table as it appears in the connector schema. Currently, this is only used in our Netsuite2 data models, whose actual table names and defined `source` table names differ slightly due to our support of both the Netsuite1 and Netsuite2 endpoints.
```

macros/union_data.sql
```diff
@@ -116,13 +118,14 @@
         {%- set relation.value=adapter.get_relation(
             database=source(default_schema, table_identifier).database,
             schema=source(default_schema, table_identifier).schema,
-            identifier=var(identifier_var, table_identifier)
+            identifier=connector_table_name_override if (connector_table_name_override and not var('use_table_name_identifer_override', false)) else var(identifier_var, connector_table_name_override if connector_table_name_override else table_identifier)
```
fivetran-joemarkiewicz, line 121: I believe I am understanding this; however, for posterity sake would you be able to provide an example for each of these conditionals and what the expected behavior would be?

fivetran-joemarkiewicz (commented):
@fivetran-jamie I just have a few comments and minor changes. Would you be able to address these. Once they are addressed I can take a final look and ensure buildkite is passing before releasing. 

Also I asked eng to remove the quickstart.yml test from this repo.

macros/union_data.sql
fivetran-jamie: Sure, let's take the account_type table in Netsuite2. The [`source`](https://github.com/fivetran/dbt_netsuite_source/blob/feature/add-union-data/models/netsuite2/src_netsuite2.yml#L16) table name is `account_type`, while the actual connector table name is `accounttype`. 

In the [_tmp](https://github.com/fivetran/dbt_netsuite_source/blob/feature/add-union-data/models/netsuite2/tmp/stg_netsuite2__account_types_tmp.sql) model, we provide `accounttype` as the `connector_table_name_override` instead of the typical `table_identifier`, which would be `account_type` in this case. we have not configured the `use_table_name_identifier_override` variable, so that remains false and the macro uses the `connector_table_name_override`. For tables that don't have special/wonky connector names, the macro uses the default `table_identifier` (example: [accounts](https://github.com/fivetran/dbt_netsuite_source/blob/feature/add-union-data/models/netsuite2/tmp/stg_netsuite2__accounts_tmp.sql))

Back to account_types -- in our integration tests, the seed file is named `netsuite2_account_type_data`. We use the `netsuite2_account_type_identifier` variable to point to this name [here](https://github.com/fivetran/dbt_netsuite_source/blob/feature/add-union-data/integration_tests/dbt_project.yml#L49). However, because the respective _tmp model has a `connector_table_name_override`, i was seeing the `no ACCOUNT_TYPE table found in your schema. Fivetran will create an empty staging model yadda yadda` log message for this and every table for which we used `connector_table_name_override`.

To ensure that we _could_ use identifier variables even for these weird tables, we now set the global `use_table_name_identifier_override` variable to True in the integration_tests/dbt_project.yml file. This means that we'll use the `connector_table_name_override` if it's provided, but _only_ if we're not overriding the table identifier completely with our own custom identifier (in this case `netsuite2_account_type_data`). 

But let's say that our seed file for `accounting_book_subsidiaries` aligns with the actual connector table name of `accountingbooksubsidiaries` (which we have configured in the [_tmp](https://github.com/fivetran/dbt_netsuite_source/blob/feature/add-union-data/models/netsuite2/tmp/stg_netsuite2__accounting_book_subsidiaries_tmp.sql) model as well), so it's `accountingbooksubsidiaries.csv`. 

The last part of the highlighted line, `var(identifier_var, connector_table_name_override if connector_table_name_override else table_identifier)` makes it so that we can use identifier variables for some tables but we don't _need_ to use them for all of the tables (despite our globally-defined variable). The package will pick up `accountingbooksubsidiaries` for the respective staging model and still pick up `netsuite2_account_type_data` for the account_type staging model. This would work as well if we had a seed file called `accounts.csv` for the accounts table, which doesn't use the `connector_table_name_override`.

PR #145, opened 2024-04-18
Feature/add lookback macro

fivetran-catfritz:
 

> - Added the new `fivetran_lookback()` macro. This generates SQL to simplify adding lookback windows to incremental models. See [README](https://github.com/fivetran/dbt_fivetran_utils/blob/releases/v0.4.latest/README.md#fivetran_lookback-source).
> - Added the new `is_databricks_sql_warehouse()` macro. This determines if Databricks destination is a SQL Warehouse or All Purpose Cluster. See [README](https://github.com/fivetran/dbt_fivetran_utils/blob/releases/v0.4.latest/README.md#is-databricks-sql-warehouse-source).

- Tested in my updates for Jira, Netsuite, and Shopify. I also made sure the check the generated compiled code for each warehouse to ensure it was generating the SQL as expected. 

N/A

- [x] Yes


integration_tests/packages.yml
```diff
@@ -2,8 +2,10 @@ packages:
   - package: fivetran/ad_reporting
     version: [">=1.0.0", "<2.0.0"]
 
-  - package: fivetran/shopify_holistic_reporting
-    version: [">=0.1.0", "<1.0.0"]
+  # - package: fivetran/shopify_holistic_reporting
+  #   version: [">=0.1.0", "<1.0.0"]
+  - git: https://github.com/fivetran/dbt_shopify_holistic_reporting.git
+    revision: feature/bump-packages
```
fivetran-joemarkiewicz, line 8: Would it be possible to swap this before release?

macros/fivetran_lookback.sql
```diff
@@ -0,0 +1,18 @@
...
+{%- endmacro %}
+
+{% macro default__fivetran_lookback(from_date, datepart, interval, safety_date='2010-01-01')  %}
+
+    {% set sql_statement %}
+        select coalesce({{ from_date }}, {{ "'" ~ safety_date ~ "'" }})
+        from {{ this }}
+    {%- endset -%}
+
+    {%- set result = dbt_utils.get_single_value(sql_statement) %}
```
fivetran-joemarkiewicz, line 14: I actually don't have a recollection of using the `dbt_utils.get_single_value()` macro before. Is this essentially doing the same thing as this [run_query](https://github.com/fivetran/dbt_salesforce/blob/fa353308688ed708ea2da5f8116f21ec8aba4989/models/salesforce/intermediate/int_salesforce__date_spine.sql#L13) operation? If so, the get_sing_value is much cleaner and probably what we should be using across our models!

fivetran-joemarkiewicz (commented):
@fivetran-catfritz this is looking great and I don't have any comments on your code changes that would require changes from your end before approval. The only thing I would request is for you to run buildkite and ensure all tests pass. Once they do, I can come in and give this an approval!

macros/fivetran_lookback.sql
fivetran-catfritz: It is slightly different execution, but the end result looks like it's practically the same. [Here](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/get_single_value.sql) is the source for `get_single_value()`. I found it by accident while reading documentation and thought I'd give it a try to make things easier.

integration_tests/packages.yml
fivetran-catfritz: Only if `dbt_shopify_holistic_reporting` is updated with the newest shopify package versions.

macros/is_databricks_sql_warehouse.sql
```diff
@@ -0,0 +1,15 @@
+{% macro is_databricks_sql_warehouse(target) %}
```
fivetran-jamie: since this is currently in use in the [Fivetran Platform package](https://github.com/fivetran/dbt_fivetran_log/blob/main/macros/is_databricks_sql_warehouse.sql), should we adjust the name to `fivetran_is_databricks_sql_warehouse` to not cause any conflicts? and then we can update the reference in Fivetran Platform to use this version of the macro (and delete the old one)

otherwise i think we'll need to add an adapter.dispatch line

@fivetran-catfritz

README.md
```diff
@@ -120,6 +122,17 @@ sources:
...
+----
+### is_databricks_sql_warehouse ([source](macros/is_databricks_sql_warehouse.sql))
+For Databricks destinations, this macro returns `true` if the Databricks target indicates it is a SQL Warehouse. It will return return `false` if it is an All-Purpose Cluster. 
+
+***Usage:**
+```yml
+    fivetran_utils.is_databricks_sql_warehouse(target)
+```
+**Args:**
+* `target` (required): Always set this equal to "target". This will pass all the current target information to the macro.
```
fivetran-jamie: is this actually necessary to have as an argument? like if it always must be `target` it might be harmful to have this as a parameter? cc @fivetran-joemarkiewicz as the macro author

README.md
```diff
@@ -466,6 +479,28 @@ from source
...
 
+----
+### fivetran_lookback ([source](macros/fivetran_lookback.sql))
+This macro takes a date expression and uses `dbt.dateadd` and backdates it by the specified interval. This is intended for use in incremental blocks to look backwards and catch late arriving records. Any date aggregate can be used as an input, but the typical usage is with the max date of the present model.
+
+**Usage:**
+```sql
+{% if is_incremental() %}
+  where date_day >= 
+    {{ netsuite.netsuite_lookback(
```
fivetran-jamie:
```suggestion
    {{ fivetran_utils.fivetran_lookback(
```

README.md
```diff
@@ -466,6 +479,28 @@ from source
...
+      datepart='day', 
+      safety_date='2010-01-01')
+    }}
+{% endif %}
+```
+**Args:**
+* `from_date`  (required): String to run against the current model. The expression used should be expected to return a single result.
+* `interval` (required): The number of units to look backwards.
+* `datepart` (required): The grain of the interval. 
+* `safety_date` (optional): This date will be used in the rare case that the `from_date` expression returns a null value. This only needs to be specified if you want to change the default value or '2010-01-01'.
```
fivetran-jamie: @fivetran-catfritz could you switch the order of `interval` and `datepart` to match the macro? the order of arguments in the macro file is `from_date, datepart, interval, safety_date`

macros/is_databricks_sql_warehouse.sql
fivetran-joemarkiewicz: YES! Let's rename to be safe and consistent

README.md
fivetran-joemarkiewicz: That's a good point. Especially since the target is always accessible regardless if it is a parameter or not. We can probably remove this as a parameter.

fivetran-catfritz: I tried two versions of the macro in my local Shopify to remove the argument and neither would compile with the following errors:
1. `{% macro is_databricks_sql_warehouse(target=target) %}`
    - Error: `Compilation Error` ... `'target' is undefined`
2. `{% macro is_databricks_sql_warehouse() %}`
    - Error: `Compilation Error`... `macro 'dbt_macro__is_databricks_sql_warehouse' takes not more than 0 argument(s)`

I wasn't surprised since I have had this issue in the past trying to create arg-less macros, and searching around I haven't found a solution. @fivetran-joemarkiewicz might this be why you opted in include the target object as an argument? @fivetran-jamie Do you have any ideas on how to get this to work?

fivetran-jamie: hm how did you test these? for the second one, did you remove the `target` argument from where you call the macro? the error message kinda sounds like an extra/un-needed argument has been provided to a macro call

README.md
```diff
@@ -466,6 +479,28 @@ from source
...
+      interval=3,
+      datepart='day', 
+      safety_date='2010-01-01')
+    }}
+{% endif %}
+```
+**Args:**
+* `from_date`  (required): String to run against the current model. The expression used should be expected to return a single result.
+* `interval` (required): The number of units to look backwards.
+* `datepart` (required): The grain of the interval.
```
fivetran-catfritz, line 501:
```suggestion
* `datepart` (required): The grain of the interval. 
* `interval` (required): The number of units to look backwards.
```

README.md
fivetran-catfritz:
```suggestion
* `datepart` (required): The grain of the interval. 
* `interval` (required): The number of units to look backwards.
* `safety_date` (optional): This date will be used in the rare case that the `from_date` expression returns a null value. This only needs to be specified if you want to change the default value from '2010-01-01'.
```

macros/is_databricks_sql_warehouse.sql
fivetran-catfritz:
```suggestion
{% macro fivetran_is_databricks_sql_warehouse(target) %}
```

README.md
fivetran-catfritz: I got it working! Updated!

README.md
fivetran-catfritz: Updated.

macros/is_databricks_sql_warehouse.sql
fivetran-catfritz: Updated.

fivetran-catfritz:
Thanks for your suggestions @fivetran-jamie ! I have updated.

fivetran-jamie (approved):

PR #152, opened 2025-08-08
fusion compatibility: test arguments

dataders:

I've added
1. re test arguments
  2. added flag to require that all test arguments are nested under config
  3. migrated the single test in the integration tests that's pertinent to the deprecation
4. `profile: 'default` so that Fusion actually parses instead of erroring about a missing profile field
5. `dbt_internal_packages/` -> `.gitignore`
6. rename path specifier keys that have been deprecated since dbt Core `1.0`


b-per:
Hi team. We can talk about it next week, but this should not be required as the latest release (that diverged from `main` some time ago) doesn't have parse warnings/issues today in Fusion.

dbt_project.yml
```diff
@@ -1,4 +1,22 @@
 name: 'fivetran_utils'
-version: '0.4.10'
+version: '0.5.0'
+profile: 'default'
 config-version: 2
-require-dbt-version: [">=1.3.0", "<2.0.0"]
+
+require-dbt-version: [">=1.10.6"]
```
dataders, line 5: @b-per maybe we still need a new version to remove the upper bound so that fusion works?


```suggestion
require-dbt-version: [">=1.3.0"]
```

dbt_project.yml
```diff
@@ -1,4 +1,22 @@
 name: 'fivetran_utils'
-version: '0.4.10'
+version: '0.5.0'
+profile: 'default'
 config-version: 2
-require-dbt-version: [">=1.3.0", "<2.0.0"]
+
```
dataders:
```suggestion
```

dbt_project.yml
b-per: Fusion doesn't look at those today (but might in the future)

dbt_project.yml
```diff
@@ -1,4 +1,21 @@
...
+test-paths: ["tests"]
+seed-paths: ["data"]
+macro-paths: ["macros"]
+snapshot-paths: ["snapshots"]
+
+
+vars:
+  fivetran_utils:
+    dbt_utils_dispatch_list:
+      - fivetran_utils
```
dataders, line 18:
```suggestion
```

PR #156, opened 2026-05-28
Update/union connections

fivetran-catfritz:
 
- 0.4.12
- New macros for MDLS support

- Buildkite and every package is tested individually

- Buildkite and every package is tested individuall

- [x] Yes


macros/fill_staging_columns.sql
```diff
@@ -1,35 +1,38 @@
...
-        {{ fivetran_utils.quote_column(column) }} as 
-        {%- if 'alias' in column %} {{ column.alias }} {% else %} {{ fivetran_utils.quote_column(column) }} {%- endif -%}
-    {%- else -%}
-        cast(null as {{ column.datatype }})
-        {%- if 'alias' in column %} as {{ column.alias }} {% else %} as {{ fivetran_utils.quote_column(column) }} {% endif -%}
-    {%- endif -%}
-    {%- if not loop.last -%} , {% endif -%}
-{% endfor %}
+    {%- if column.name|lower in source_column_names %}
+        {%- if using_source_casing %}
```
fivetran-joemarkiewicz, line 8: For this iteration we need it to be `not`. The prior approach was to have the default value of this variable be `true` and we set it to `false` to activate the new behavior. We can switch this back when we change the variable name.
```suggestion
        {%- if not using_source_casing %}
```

macros/fill_staging_columns.sql
```diff
@@ -1,35 +1,38 @@
...
+            {%- set column_alias = column.alias if 'alias' in column else column.name %}
+            {{ adapter.quote(column.name) }} as {{ adapter.quote(column_alias|upper if target.warehouse == 'snowflake' else column_alias) }}
```
fivetran-joemarkiewicz: it's possible target.warehouse isn't always specified. target.type would be better here.
```suggestion
            {{ adapter.quote(column.name) }} as {{ adapter.quote(column_alias|upper if target.type == 'snowflake' else column_alias) }}
```

macros/fill_staging_columns.sql
fivetran-catfritz: Ahh yes thank you updated!

macros/fill_staging_columns.sql
fivetran-catfritz: Yeah that's a mistake. Fixed!

macros/union_connections.sql
```diff
@@ -0,0 +1,94 @@
...
+    -- ** Values passed to adapter.get_relation:
+    {{ '-- full-identifier_var: ' ~ identifier_var }}
+    {{ '-- database: ' ~ database }}                            
+    {{ '-- schema: ' ~ schema }}
+    {{ '-- identifier: ' ~ identifier ~ '\n' }}
+
+    {% if relation is not none -%}
+        select
+            {{ dbt_utils.star(from=source(single_source_name, single_table_name)) }}
+            , '{{ var("zendesk_database", target.database) }}' || '.'|| '{{ var("zendesk_schema", "zendesk") }}' as _dbt_source_relation
```
fivetran-joemarkiewicz: This needs to change this to not be zendesk specific.

macros/apply_source_relation.sql
```diff
@@ -0,0 +1,11 @@
...
+
+{{ adapter.dispatch('apply_source_relation', 'fivetran_utils') () }}
+
+{%- endmacro %}
+
+{% macro default__apply_source_relation() -%}
+
+, _dbt_source_relation as source_relation
+
+{%- endmacro %}
```
fivetran-joemarkiewicz: This works for the _newer_ union data packages. However, for ones that must retain backwards compatibility, this doesn't work. Proposing these changes to ensure this works in backwards compatibile situations.


```suggestion
{% macro apply_source_relation(package_name, use_package_prefix=true) -%}

{{ adapter.dispatch('apply_source_relation', 'fivetran_utils') (package_name, use_package_prefix) }}

{%- endmacro %}

{% macro default__apply_source_relation(package_name, use_package_prefix=true) -%}

{% set sources_var = package_name ~ '_sources' %}
{% set database_var = package_name ~ '_database' %}
{% set schema_var = package_name ~ '_schema' %}

{% if use_package_prefix %}
    {% set union_schemas_var = package_name ~ '_union_schemas' %}
    {% set union_databases_var = package_name ~ '_union_databases' %}
{% else %}
    {% set union_schemas_var = 'union_schemas' %}
    {% set union_databases_var = 'union_databases' %}
{% endif %}

{% if var(sources_var, []) != [] %}
, _dbt_source_relation as source_relation
{% elif var(union_schemas_var, []) != [] or var(union_databases_var, []) != [] %}
{{ fivetran_utils.source_relation() }}
{% else %}
{% set database = var(database_var, target.database) %}
{% set schema = var(schema_var, package_name) %}
, '{{ database ~ "." ~ schema }}' as source_relation
{% endif %}

{%- endmacro %}
```

fivetran-catfritz: Ahh yes good call. My only suggestion would be for the condition to check the length of the list instead of a match, since that's been less brittle. This is how I did it in the new `partition_by_source_relation` for example, and also how we do it in `union_connections`.

```
var(package_name ~ '_union_schemas', []) | length > 0
```

macros/partition_by_source_relation.sql
```diff
@@ -0,0 +1,16 @@
...
+{%- set is_unioning = var(package_name ~ '_union_schemas', [])|length > 1 or var(package_name ~ '_union_databases', [])|length > 1 or var(package_name ~ '_sources', [])|length > 1 -%}
+{%- set prefix = '' if alias is none else alias ~ '.' -%}
+
+{%- if has_other_partitions == 'no' -%}
+    {{- 'partition by ' ~ prefix ~ 'source_relation' if is_unioning else '' -}}
+{%- else -%}
+    {{- ', ' ~ prefix ~ 'source_relation' if is_unioning else '' -}}
+{%- endif -%}
+
+{%- endmacro %}
```
fivetran-joemarkiewicz: Similar to the apply_source_relation suggestion. This ensures backwards compatibility.
```suggestion
{% macro partition_by_source_relation(package_name, has_other_partitions='yes', alias=None, package_prefix_union_variable=true) %}
    {{ return(adapter.dispatch('partition_by_source_relation', 'fivetran_utils')(package_name, has_other_partitions, alias, package_prefix_union_variable)) }}
{% endmacro %}

{% macro default__partition_by_source_relation(package_name, has_other_partitions='yes', alias=None, package_prefix_union_variable=true) -%}

{%- if package_prefix_union_variable %}
    {%- set union_schemas_var = package_name ~ '_union_schemas' -%}
    {%- set union_databases_var = package_name ~ '_union_databases' -%}
{%- else %}
    {%- set union_schemas_var = 'union_schemas' -%}
    {%- set union_databases_var = 'union_databases' -%}
{%- endif -%}

{%- set is_unioning = var(union_schemas_var, [])|length > 1 or var(union_databases_var, [])|length > 1 or var(package_name ~ '_sources', [])|length > 1 -%}
{%- set prefix = '' if alias is none else alias ~ '.' -%}

{%- if has_other_partitions == 'no' -%}
    {{- 'partition by ' ~ prefix ~ 'source_relation' if is_unioning else '' -}}
{%- else -%}
    {{- ', ' ~ prefix ~ 'source_relation' if is_unioning else '' -}}
{%- endif -%}

{%- endmacro %}
```

macros/fill_staging_columns.sql
```diff
@@ -1,35 +1,38 @@
-{% macro fill_staging_columns(source_columns, staging_columns) -%}
+{% macro fill_staging_columns(source_columns, staging_columns) %}
 
 {%- set source_column_names = source_columns|map(attribute='name')|map('lower')|list -%}
+{%- set using_source_casing = var('using_deliberate_quoting', true) -%}
```
fivetran-joemarkiewicz: We can swap this back

fivetran-joemarkiewicz: Default will be false

macros/fill_staging_columns.sql
```diff
@@ -1,35 +1,38 @@
...
-        {{ fivetran_utils.quote_column(column) }} as 
-        {%- if 'alias' in column %} {{ column.alias }} {% else %} {{ fivetran_utils.quote_column(column) }} {%- endif -%}
-    {%- else -%}
-        cast(null as {{ column.datatype }})
-        {%- if 'alias' in column %} as {{ column.alias }} {% else %} as {{ fivetran_utils.quote_column(column) }} {% endif -%}
-    {%- endif -%}
-    {%- if not loop.last -%} , {% endif -%}
-{% endfor %}
+    {%- if column.name|lower in source_column_names %}
+        {%- if not using_source_casing %}
```
fivetran-joemarkiewicz: Change this to remove `not`

macros/fill_staging_columns.sql
```diff
@@ -1,35 +1,38 @@
...
+            {%- set column_alias = column.alias if 'alias' in column else column.name %}
+            {{ adapter.quote(column.name|lower) }} as {{ adapter.quote(column_alias|upper if target.type == 'snowflake' else column_alias) }}
```
fivetran-joemarkiewicz: Remove the `lower` filter

macros/fill_staging_columns.sql
fivetran-catfritz:
```suggestion
{%- set using_source_casing = var('fivetran_using_source_casing', false) -%}
```

macros/fill_staging_columns.sql
fivetran-catfritz:
```suggestion
        {%- if using_source_casing %}
```

macros/fill_staging_columns.sql
fivetran-catfritz:
```suggestion
            {{ adapter.quote(column.name) }} as {{ adapter.quote(column_alias|upper if target.type == 'snowflake' else column_alias) }}
```

macros/apply_source_relation.sql
```diff
@@ -0,0 +1,31 @@
...
+{% endif %}
+
+{% if var(sources_var, []) | length > 0 %}
+, _dbt_source_relation as source_relation
+{% elif var(union_schemas_var, []) | length > 0 or var(union_databases_var, []) | length > 0 %}
+{{ fivetran_utils.source_relation(union_schema_variable=union_schemas_var, union_database_variable=union_databases_var) }}
+{% else %}
+{% set database = var(database_var, target.database) %}
+{% set schema = var(schema_var, package_name) %}
+, '{{ database ~ "." ~ schema }}' as source_relation
```
fivetran-catfritz: @fivetran-joemarkiewicz I was getting the error I screenshotted below (in bk and locally) for redshift. This suggestion fixed it. Just wanted to run it by you before I pushed to this branch.

```suggestion
, cast('{{ database ~ "." ~ schema }}' as {{ dbt.type_string() }}) as source_relation
```

- Before
Screenshot 2026-06-08 at 7 55 58 PM

- After
Screenshot 2026-06-08 at 7 58 47 PM

README.md
```diff
@@ -504,6 +501,113 @@ It should be added to all non-tmp staging models when using the `union_data` mac
 * `union_schema_variable` (optional): The name of the union schema variable. By default the macro will look for `union_schemas`.
 * `union_database_variable` (optional): The name of the union database variable. By default the macro will look for `union_databases`.
 
+----
+### apply_source_relation ([source](macros/apply_source_relation.sql))
+This macro generates the `source_relation` column in non-tmp staging models. It automatically selects the right approach based on how a package is configured for unioning:
+- If `{package_name}_sources` is set (new-style `union_connections`): passes through `_dbt_source_relation as source_relation`.
+- If `{package_name}_union_schemas` or `{package_name}_union_databases` is set: calls `source_relation()` with the appropriate variable names.
+- If none of the above: generates a static `database.schema` string cast as `source_relation`.
```
fivetran-joemarkiewicz: Small edit to clarify the source_relation macro is a Fivetran Utils one.
```suggestion
- If `{package_name}_union_schemas` or `{package_name}_union_databases` is set: calls `fivetran_utils.source_relation()` with the appropriate variable names.
- If none of the above: generates a static `database.schema` string cast as `source_relation`.
```

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
...
+  - [`union_relations_custom`](macros/union_relations_custom.sql): A Fivetran-specific fork of `dbt_utils.union_relations` that sets `_dbt_source_relation` to `database.schema` only (excluding the table name) to align with Fivetran package conventions.
+  - [`apply_source_relation`](macros/apply_source_relation.sql): Derives and appends the `source_relation` column in staging models.
+  - [`partition_by_source_relation`](macros/partition_by_source_relation.sql): Returns a `PARTITION BY source_relation` clause (or `, source_relation`) only when more than one source is detected.
+- Updates the `fill_staging_columns` macro to support a new `fivetran_using_source_casing` variable (default `false`). When set to `true`, the macro quotes lowercase column names as defined in the package's `get_*_columns` macros, with Snowflake targets uppercasing the alias to match Snowflake identifier conventions in downstream models. This was added to support Fivetran's MDLs feature where Polaris lowercases column names in Snowflake. See the [DECISIONLOG](DECISIONLOG.md) for more details.
```
fivetran-joemarkiewicz:
```suggestion
- Updates the `fill_staging_columns` macro to support a new `fivetran_using_source_casing` variable (default `false`). When set to `true`, the macro quotes lowercase column names as defined in the package's `get_*_columns` macros. On Snowflake targets, aliases are uppercased to match Snowflake's identifier conventions in downstream models. This was added to support Fivetran's Managed Data Lake Service, where column names are stored as lowercase in the Iceberg layer and must be quoted to be referenced correctly in Snowflake. See the [DECISIONLOG](DECISIONLOG.md) for more details.
```

DECISIONLOG.md
```diff
@@ -0,0 +1,9 @@
+# Decision Log
+
+## Why `fivetran_using_source_casing` quotes lowercase column names instead of preserving source casing
+
+The `fivetran_using_source_casing` variable in `fill_staging_columns` was added to support Fivetran's Managed Data Lakes (MDLs) feature, where the Polaris engine lowercases column names in Snowflake destinations. Because Snowflake treats unquoted identifiers as uppercase by default, column names written in lowercase by Polaris must be quoted to be referenced correctly.
```
fivetran-joemarkiewicz:
```suggestion
The `fivetran_using_source_casing` variable in `fill_staging_columns` was added to support Fivetran's Managed Data Lake Service, where column names are stored as lowercase in the Iceberg layer. When querying these tables from Snowflake via the Polaris catalog, which treats unquoted identifiers as uppercase by default, those lowercase column names must be quoted to be referenced correctly.
```

macros/union_relations_custom.sql
```diff
@@ -0,0 +1,130 @@
...
+            {%- do all_includes.append(inc | lower) -%}
+        {%- endfor -%}
+    {%- endif -%}
+
+    {%- for relation in relations -%}
+
+        {%- do relation_columns.update({relation: []}) -%}
+
+        {%- do dbt_utils._is_relation(relation, 'zendesk_union_relations') -%}
+        {%- do dbt_utils._is_ephemeral(relation, 'zendesk_union_relations') -%}
```
fivetran-jamie: I'm not really sure what these lines do, but should it be `union_relations_custom` instead of `zendesk_union_relations`?

README.md
```diff
@@ -577,6 +681,31 @@ sources:
...
+This macro is adapted from dbt_utils.union_relations() with a slight deviation, allowing for the `_dbt_source_relation` field to return only the `database.schema` as opposed to the `database.schema.table`. 
+Unions together an array of [Relations](https://docs.getdbt.com/docs/writing-code-in-dbt/class-reference/#relation),
+even when columns have differing orders in each Relation, and/or some columns are
+missing from some relations. Any columns exclusive to a subset of these
+relations will be filled with `null` where not present. An new column
+(`_dbt_source_relation`) is also added to indicate the source for each record.
+
+**Usage:**
+```sql
+{{ dbt_utils.union_relations(
```
fivetran-jamie:
```suggestion
{{ fivetran_utils.union_relations_custom(
```

README.md
```diff
@@ -577,6 +681,31 @@ sources:
...
+    exclude=["_loaded_at"]
+) }}
+```
+**Args:**
+* `relations`          (required): An array of [Relations](https://docs.getdbt.com/docs/writing-code-in-dbt/class-reference/#relation).
+* `aliases`            (optional): An override of the relation identifier. This argument should be populated with the overwritten alias for the relation. If not populated `relations` will be the default.
+* `exclude`            (optional): A list of column names that should be excluded from the final query.
+* `include`            (optional): A list of column names that should be included in the final query. Note the `include` and `exclude` parameters are mutually exclusive.
+* `column_override`    (optional): A dictionary of explicit column type overrides, e.g. `{"some_field": "varchar(100)"}`.``
+* `source_column_name` (optional): The name of the column that records the source of this row. By default this argument is set to `none`.
```
fivetran-jamie: I think the default is `_dbt_source_relation`

README.md
```diff
@@ -646,7 +775,7 @@ packages:
       version: [">=1.0.0", "<2.0.0"]
 ```
 
-# 🙌 How is this package maintained and can I contribute?
+# How is this package maintained and can I contribute?
```
fivetran-jamie, line 780: Could you update the links on line 780 to point here instead of Jira

fivetran-jamie (commented):
This is looking and running great. Couple of minor comments

macros/union_relations_custom.sql
fivetran-joemarkiewicz: I noticed the same. I believe it should be `union_relations_custom`

macros/union_connections.sql
```diff
@@ -0,0 +1,94 @@
...
+    -- ** Values passed to adapter.get_relation:
+    {{ '-- full-identifier_var: ' ~ identifier_var }}
+    {{ '-- database: ' ~ database }}                            
+    {{ '-- schema: ' ~ schema }}
+    {{ '-- identifier: ' ~ identifier ~ '\n' }}
+
+    {% if relation is not none -%}
+        select
+            {{ dbt_utils.star(from=source(single_source_name, single_table_name)) }}
+            , '{{ database ~ "." ~ schema }}' as _dbt_source_relation
```
fivetran-joemarkiewicz: This addition is a change from the implementation seen in the existing packages (eg. [zendesk](https://github.com/fivetran/dbt_zendesk/blob/84499df2e775fce3a683ec4ad627a51c95ab59b3/macros/staging/union/union_zendesk_connections.sql#L55-L57), [recharge](https://github.com/fivetran/dbt_recharge/blob/069ccb075f9549e8bd689b469d029689f2cbc9d9/macros/union/union_connections.sql#L80-L82), [netsuite](https://github.com/fivetran/dbt_netsuite/blob/6d7d1a18ee9ba2d2ca3ccf1ce904b6ed60c35f30/macros/union/netsuite_union_connections.sql#L61-L63)) for non-unioning situations. For example, when running a job on Zendesk without unioning on this branch I see the `_dbt_source_relation` field in the tmp model. However, when running on the latest version of Fivetran Utils, I don't see that field.

Is this addition a requirement? When looking at the apply_source_relation macro it looks like when not unioning we ultimately create the `source_relation` field from [scratch](https://github.com/fivetran/dbt_fivetran_utils/blob/a6d93c52abd339a50d8b01441da4b50d5177de1e/macros/apply_source_relation.sql#L25-L29), so this _dbt_source_relation is never used. 

If this isn't required, let's remove it as it's not being used by the `apply_source_relation` macro in the downstream staging model. Further, if left this will result in a schema change for the tmp models of existing jobs not using the union feature.

fivetran-joemarkiewicz (changes requested):
A few wording update suggestions and some code questions to be addressed before approval.

macros/union_connections.sql
fivetran-catfritz:
```suggestion
            {{ dbt_utils.star(from=source(single_source_name, single_table_name)) }}
```

macros/apply_source_relation.sql
```diff
@@ -0,0 +1,31 @@
...
+
+{{ adapter.dispatch('apply_source_relation', 'fivetran_utils') (package_name, use_package_prefix) }}
+
+{%- endmacro %}
+
+{% macro default__apply_source_relation(package_name, use_package_prefix=true) -%}
+
+{% set sources_var = package_name ~ '_sources' %}
+{% set database_var = package_name ~ '_database' %}
+{% set schema_var = package_name ~ '_schema' %}
```
fivetran-catfritz, line 11: @fivetran-joemarkiewicz @fivetran-jamie I just realized we need to account for packages like `pinterest_ads`, where the `package_name` is `pinterest_ads` but the schema var is [`pinterest_schema`](https://github.com/fivetran/dbt_pinterest/blob/main/integration_tests/dbt_project.yml#L7).

My question for you both: should we handle that mismatch here, or should we chain the schema and database vars in Pinterest and any other packages that don’t follow the naming convention, so we solve this more broadly? And also since we're already updating those packages? 

For example, in Pinterest, where we currently use `var('pinterest_schema', default)`, we could update it to something like:

`var('pinterest_ads_schema', var('pinterest_schema', default))`

That way, it supports either var name. Thoughts?

fivetran-joemarkiewicz: I agree with your suggestion. Let's solve this more broadly. Your suggestion looks good! Let's confirm this works (chaining together the variable default values) and if it does then let's take this approach.

macros/union_relations_custom.sql
fivetran-catfritz:
```suggestion
        {%- do dbt_utils._is_relation(relation, 'union_relations_custom') -%}
        {%- do dbt_utils._is_ephemeral(relation, 'union_relations_custom') -%}
```

README.md
fivetran-catfritz: Yes, thanks. Updated.
```suggestion
* `source_column_name` (optional): The name of the column that records the source of this row. By default this argument is set to `_dbt_source_relation`.
```

README.md
```diff
@@ -646,7 +775,7 @@ packages:
...
 ## Package Maintenance
 The Fivetran team maintaining this package **only** maintains the latest version of the package. We highly recommend you stay consistent with the [latest version](https://hub.getdbt.com/fivetran/jira/latest/) of the package and refer to the [CHANGELOG](https://github.com/fivetran/dbt_jira/blob/main/CHANGELOG.md) and release notes for more information on changes across versions.
```
fivetran-catfritz:
```suggestion
The Fivetran team maintaining this package **only** maintains the latest version of the package. We highly recommend you stay consistent with the [latest version](https://hub.getdbt.com/fivetran/fivetran_utils/latest/) of the package and refer to the [CHANGELOG](https://github.com/fivetran/dbt_fivetran_utils/blob/releases/v0.4.latest/CHANGELOG.md) and release notes for more information on changes across versions.
```

.buildkite/pipeline.yml
fivetran-catfritz:
```suggestion
      bash .buildkite/scripts/run_standard_models.sh databricks facebook_pages instagram_business twitter_organic youtube_analytics
```
## Removed linkedin_pages as it is not databricks compatible at the moment.

macros/apply_source_relation.sql
fivetran-catfritz: Updated.

macros/union_relations_custom.sql
fivetran-catfritz: Updated!

README.md
fivetran-catfritz: Updated.

macros/union_connections.sql
fivetran-catfritz: Removed.

macros/union_connections.sql
```diff
@@ -0,0 +1,93 @@
+{% macro union_connections(connection_dictionary, single_source_name, single_table_name, default_identifier=single_table_name) %}
+
+{{ adapter.dispatch('union_connections', 'fivetran_utils') (connection_dictionary, single_source_name, single_table_name, default_identifier) }}
+
+{%- endmacro %}
+
+{% macro default__union_connections(connection_dictionary, single_source_name, single_table_name, default_identifier=single_table_name) %}
+
+{%- set exception_warning = "\n\nPlease be aware: The " ~ single_source_name|upper ~ "." ~ single_table_name|upper ~ " table was not found in your schema(s). The Fivetran Data Model will create a completely empty staging model as to not break downstream transformations. To turn off these warnings, set the `fivetran__remove_empty_table_warnings` variable to TRUE (see https://github.com/fivetran/dbt_fivetran_utils/tree/releases/v0.4.latest#union_data-source for details).\n"%}
```
fivetran-joemarkiewicz: We also need to update the README for this macro to highlight the empty table behavior and use of the `fivetran__remove_empty_table_warnings` variable. We can probably leverage the same wording as used in the union_data documentation.
```suggestion
{%- set exception_warning = "\n\nPlease be aware: The " ~ single_source_name|upper ~ "." ~ single_table_name|upper ~ " table was not found in your schema(s). The Fivetran Data Model will create a completely empty staging model as to not break downstream transformations. To turn off these warnings, set the `fivetran__remove_empty_table_warnings` variable to TRUE (see https://github.com/fivetran/dbt_fivetran_utils/tree/releases/v0.4.latest#union_connections-source for details).\n"%}
```

README.md
```diff
@@ -504,6 +501,113 @@ It should be added to all non-tmp staging models when using the `union_data` mac
 * `union_schema_variable` (optional): The name of the union schema variable. By default the macro will look for `union_schemas`.
 * `union_database_variable` (optional): The name of the union database variable. By default the macro will look for `union_databases`.
 
+----
+### apply_source_relation ([source](macros/apply_source_relation.sql))
```
fivetran-joemarkiewicz, line 505: Same comment to apply the empty table behavior documentation here.

fivetran-joemarkiewicz (approved):
LGTM with two small changes to be applied before merge/release.

README.md
```diff
@@ -82,18 +70,23 @@ dispatch:
     - [persist\_pass\_through\_columns (source)](#persist_pass_through_columns-source)
     - [remove\_prefix\_from\_columns (source)](#remove_prefix_from_columns-source)
     - [source\_relation (source)](#source_relation-source)
+    - [apply\_source\_relation (source)](#apply_source_relation-source)
+    - [partition\_by\_source\_relation (source)](#partition_by_source_relation-source)
+    - [union\_connections (source)](#union_connections-source)
+      - [Union Connections Defined Sources Configuration](#union-connections-defined-sources-configuration)
```
fivetran-jamie:
```suggestion
      - [Union Connections Defined Sources Configuration](#optional-union-connections-defined-sources-configuration)
```
Link is currently broken

README.md
```diff
@@ -461,8 +453,13 @@ select
...
+* `source_columns` (required): Calls [get_columns_in_relation](https://docs.getdbt.com/reference/dbt-jinja-functions/adapter/#get_columns_in_relation) and requires a `ref()` or `source()` pointing to the `_tmp` staging model.
+* `staging_columns` (required): The list of expected columns, typically returned by the package's `get_<table>_columns()` macro.
+
+**Variables:**
+* `fivetran_using_source_casing` (optional): Boolean variable, defaults to `false`. When `true`, the macro quotes lowercase column names as defined in the package's `get_*_columns` macros. For Snowflake targets, the alias is uppercased to match Snowflake identifier conventions.
+
+> **Important:** `fivetran_using_source_casing` was added specifically to support Fivetran's MDLs feature where the Polaris engine lowercases column names in Snowflake destinations. It does not generically preserve source casing — it mianly handles the lowercase-in-Snowflake case. If Polaris's behavior changes in the future, a different approach may be required. See the [DECISIONLOG](DECISIONLOG.md) for full details.
```
fivetran-jamie:
```suggestion
> **Important:** `fivetran_using_source_casing` was added specifically to support Fivetran's MDLs feature where the Polaris engine lowercases column names in Snowflake destinations. It does not generically preserve source casing — it mainly handles the lowercase-in-Snowflake case. If Polaris's behavior changes in the future, a different approach may be required. See the [DECISIONLOG](DECISIONLOG.md) for full details.
```

README.md
```diff
@@ -577,6 +681,31 @@ sources:
...
+    relations=[ref('my_model'), source('my_source', 'my_table')],
+    exclude=["_loaded_at"]
+) }}
+```
+**Args:**
+* `relations`          (required): An array of [Relations](https://docs.getdbt.com/docs/writing-code-in-dbt/class-reference/#relation).
+* `aliases`            (optional): An override of the relation identifier. This argument should be populated with the overwritten alias for the relation. If not populated `relations` will be the default.
+* `exclude`            (optional): A list of column names that should be excluded from the final query.
+* `include`            (optional): A list of column names that should be included in the final query. Note the `include` and `exclude` parameters are mutually exclusive.
+* `column_override`    (optional): A dictionary of explicit column type overrides, e.g. `{"some_field": "varchar(100)"}`.``
```
fivetran-jamie:
```suggestion
* `column_override`    (optional): A dictionary of explicit column type overrides, e.g. `{"some_field": "varchar(100)"}`.
```

This is present on line 729 for `union_relations` as well

fivetran-jamie (approved):
Approved with a few cosmetic README suggestions

README.md
```diff
@@ -504,6 +501,113 @@ It should be added to all non-tmp staging models when using the `union_data` mac
...
+```
+**Args:**
+* `package_name` (required): The name of the package. Used to derive the union variable names.
+* `has_other_partitions` (optional): Defaults to `'yes'`. When `'yes'`, prepends `, source_relation` to add to an existing partition. When `'no'`, generates the full `partition by source_relation` clause.
+* `alias` (optional): A CTE or relation alias to prefix `source_relation` (e.g., produces `alias.source_relation`).
+* `package_prefix_union_variable` (optional): Boolean, defaults to `true`. When `true`, prefixes union variable names with `package_name`. When `false`, uses generic names (`union_schemas`, `union_databases`).
+
+----
+### union_connections ([source](macros/union_connections.sql))
+This macro unions identically structured tables across multiple Fivetran connections of the same type. It reads from a connection dictionary variable (e.g. `jira_sources`) listing each connection's `database`, `schema`, and `name` (optional), and falls back to a single-connection query if the variable is empty. If a source table is not found, the macro returns an empty table with a single `_dbt_source_relation` column and outputs a compiler warning.
```
fivetran-catfritz:
```suggestion
This macro unions identically structured tables across multiple Fivetran connections of the same type. It reads from a connection dictionary variable (e.g. `jira_sources`) listing each connection's `database`, `schema`, and `name` (optional), and falls back to a single-connection query if the variable is empty.

If the source table is not found, `union_connections` will return a **completely** empty table (ie `limit 0`) with just one string column (`_dbt_source_relation`). A compiler warning message will be output, highlighting that the expected source table was not found and its respective staging model is empty. The compiler warning can be turned off by setting the `fivetran__remove_empty_table_warnings` variable to `True`.
```

